PR #42, opened 2015-01-01
Multiple pages

will-moore:
Ready for testing....


will-moore:
Not specifically related to multiple pages, but with a large number of images, the 'metadata' PDF page gets longer than a single page.

PR #43, opened 2015-01-06
Align magnification of selected panels

will-moore:
Fixes https://github.com/will-moore/figure/issues/34/

To test:
 - Select different sized panels with different zoom levels and pixel sizes.
 - Add similar sized scale bars to each, E.g. 5 microns.
 - Select panels and choose toolbar 'Align Sizes' > 'Align Magnification'.
 - All panels should be resized so that scale bars are same size in them all, using top-right panel as the reference.


joshmoore:
```
npm ERR! Error: No compatible version found: grunt-contrib-concat@'^0.5.0'
```

PR #52, opened 2015-01-28
Handle unicode characters by encoding/decoding

will-moore:
Need to be able to encode various unicode characters when we save files, and decode them when opening files.
Saved files need to use strings for file names (not unicode) so we also need to use the figure name encoded in the json for display etc, not the name of the file on disk.

### DIFF
--- a/views.py
+++ b/views.py
@@ -158,7 +158,11 @@ def save_web_figure(request, conn=None, **kwargs):
     if figureJSON is None:
         return HttpResponse("No 'figureJSON' in POST")
     # See https://github.com/will-moore/figure/issues/16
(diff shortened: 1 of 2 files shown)


views.py
```diff
@@ -158,7 +158,11 @@ def save_web_figure(request, conn=None, **kwargs):
     if figureJSON is None:
         return HttpResponse("No 'figureJSON' in POST")
     # See https://github.com/will-moore/figure/issues/16
-    figureJSON = unicodedata.normalize('NFKD', figureJSON).encode('ascii','ignore')
+    print type(figureJSON)
```
joshmoore: `print` heads up

PR #69, opened 2015-04-12
Crop dialog fixes

will-moore:
Addresses several points raised during OMERO.figure testing of the crop dialog:
To test:

 - Crop dialog now has a short message instructing user to drag to select a region or pick ROI.
 - If no ROIs on the image, show a message to indicate it.
 - Dialog "OK" is disabled when the dialog is first shown and is only enabled when an ROI is picked or a region is selected.
 - ROIs can be selected by clicking on the whole ROI row (don't need to click thumbnail) and this is indicated by mouse pointer and hover colour change.

### DIFF
--- a/static/figure/css/figure.css
+++ b/static/figure/css/figure.css
@@ -701,6 +701,14 @@
         z-index: 10;
     }
 
+    .roiPickMe {
+        cursor: pointer;
+    }
+    /* On hover, use selected blue color */
+    .roiPickMe:hover {
+        background-color: rgb(190, 212, 253);
+    }
+
 
     /*  color-picker  */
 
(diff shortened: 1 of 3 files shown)


will-moore:
@pwalczysko That should address a bunch of your comments from last week's testing. Should be in tomorrow's build.

pwalczysko:
As discussed, two problems here:

  - OK button never gets enabled now, neither with pre-existing nor with new regions drawn
  - the message ``no regions present`` or similar is missing in case no pre-existing regions are there

pwalczysko:
Seems that this https://github.com/will-moore/figure/pull/69#issuecomment-92889621 is actually working on trout merge, but not on trout latest. 
On the other hand, today the adding of new images does not work on trout merge.
I suggest we retest tomorrow on trout merge to see where we are.

pwalczysko:
I have relisted this for tomorrwo under Others

pwalczysko:
Works fine, except one thing:

  - bring up the Crop to ROI dialog
  - select a one-point ROI (just click, do not drag)
  - click OK
  - the pane is now of height 0, which causes it to be invisible on the canvas, the resize button next to the crop button does not help, only Undo helps

will-moore:
Last comments addressed by https://github.com/will-moore/figure/pull/75

pwalczysko:
Works as expected.

PR #75, opened 2015-04-15
Zero sized crop region

will-moore:
This PR ensures that the "OK" submit button on the crop dialog is not enabled when the region selected is zero or 1 pixel width or height.
To test:

 - Open crop dialog, single click on image to add region that has zero width & height.
 - OK button should be disabled.
 - With a larger region, it is possible to drag the region outside of the area of the image.
 - This should also disable the OK button.


pwalczysko:
Works as expected.

PR #76, opened 2015-04-16
Show 'Pixel Size NOT SET' instead of 'null'

will-moore:
Small fix to display of pixel size when it's not set.
Shows 'NOT SET' instead of 'null'.

screen shot 2015-04-16 at 23 00 43


pwalczysko:
Works as expected.

PR #81, opened 2015-04-20
Fix potential bug with adding big images to small paper

will-moore:
See https://github.com/will-moore/figure/issues/37
If a large image is added to a much smaller paper then we could end up with a negative scale number because the spacer around the image (image_width/20) was first subtracted from paper width before scaling. This is fixed by the first commit below.


will-moore:
Now works when adding an 8k * 8k image to the default A4 page.

PR #82, opened 2015-04-27
1 1 followup fixes

will-moore:
A bunch of follow-ups that came from OMERO.figure 1.1.0 release testing.

To test:
 - When prompted to Save a newly created file when using browser back button, default name is same as elsewhere (not 'unsaved').
 - Menu: File > Add Figure Legend  to add or edit a figure legend.
 - Exporting 'with images' adds a README.txt file to the zip to explain what images are in the directories.
 - Scalebars are 50% thicker, both in the web UI and exported figures.
 - Legend formatting dialog link doesn't open new page etc.
 - Clicking figure legend when collapsed expands it. Clicking when expanded starts edit.
 - Clicking links within figure legend opens in new tab/window.
 - Crop dialog should add the current panel region as initial crop selection (if zoomed in at-all).
 - Export without saving works.

### DIFF
--- a/scripts/omero/figure_scripts/Figure_To_Pdf.py
+++ b/scripts/omero/figure_scripts/Figure_To_Pdf.py
@@ -56,6 +56,24 @@ ORIGINAL_DIR = "1_originals"
 RESAMPLED_DIR = "2_pre_resampled"
 FINAL_DIR = "3_final"
 
+README_TXT = """These folders contain images used in the creation
+of the figure. Each folder contains one image per figure panel,
+with images numbered according to the order they were added to
+the figure. The numbered folders represent the sequence of
+processing steps:
+
+ - 1_originals: This contains the full-sized and un-cropped images that are
+   rendered by OMERO according to your chosen rendering settings.
+
+ - 2_pre_resampled: This folder will only contain those images that are
+   resampled in order to match the export figure resolution. This will be
+   all panels for export of TIFF figures, or individual panels that have
+   a 'dpi' set for export of PDF figures.
+
+ - 3_final: These are the image panels that are inserted into the
+   final figure, saved following any cropping, rotation and resampling steps.
+"""
+
 
 def compress(target, base):
     """
@@ -196,6 +214,7 @@ class FigureExport(object):
                 for d in (ORIGINAL_DIR, RESAMPLED_DIR, FINAL_DIR):
                     imgDir = os.path.join(zipDir, d)
                     os.mkdir(imgDir)
+                self.addReadMeFile()
 
         # Create the figure file(s)
         self.createFigure()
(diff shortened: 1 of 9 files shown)


pwalczysko:
One minor thing: the tooltip over the "Collapse legend" button in the top right corner of a legend still says "Collapse legend" even when the legend is collapsed already - it shoiuld say "Expand legen" instead in this situation, because this is what it does.

pwalczysko:
The export script is missing; again on trout merge. Contacted Seb, will come up with a correct strategy there. We probably need a more generic, robust solution.

pwalczysko:
Following points were tested already and passed the test.
```
 - When prompted to Save a newly created file when using browser back button, default name is same as elsewhere (not 'unsaved').
 - Menu: File > Add Figure Legend to add or edit a figure legend.
 - Legend formatting dialog link doesn't open new page etc.
 - Clicking figure legend when collapsed expands it. Clicking when expanded starts edit.
 - Clicking links within figure legend opens in new tab/window.
```

pwalczysko:
The rest is still ToDo - note that the export script is a blocker for testing two of the resting points.

pwalczysko:
``Export without saving works.``  - This point passed the test.

pwalczysko:
Bug: Error on export of SAVED figure - see https://trout.openmicroscopy.org/merge/figure/file/13648 - login as user-4. 

Error:
```
  File "./script", line 1210, in <module>
    runScript()
  File "./script", line 1197, in runScript
    fileAnnotation = export_figure(conn, scriptParams)
  File "./script", line 1154, in export_figure
    return figExport.buildFigure()
  File "./script", line 229, in buildFigure
    groupId = self.conn.getObject("Image", id1).getDetails().group.id.val
AttributeError: 'NoneType' object has no attribute 'getDetails'
```

pwalczysko:
I still think that the ``pre-resampled`` folder is not explained correctly in README:

```
- 2_pre_resampled: This folder will only contain those images that are
   resampled in order to match the export figure resolution. This will be
   all panels for export of TIFF figures, or individual panels that have
   a 'dpi' set for export of PDF figures.
```
This text suggests that any panel which has a set dpi will be having an image in the ``pre-resampled`` folder. This is not true - it seems you have to set a dpi AND rotate to get the image into the second (pre-resampled) folder.

pwalczysko:
``Crop dialog should add the current panel region as initial crop selection (if zoomed in at-all).``
This feature needs some further work - the problem is that you cannot click OK in such case - I understand that this is done in line with my previous comments about being able to click OK prematurely.
But in this very case, you have the region already pre-selected, so you should be able to click OK button.

pwalczysko:
In summarey:

1 RFE : Readme https://github.com/will-moore/figure/pull/82#issuecomment-97879712
1 slight bug suspicion (possibly OK, just maybe look at the output please) https://github.com/will-moore/figure/pull/82#issuecomment-97872518
1 major workflow issue with new feature https://github.com/will-moore/figure/pull/82#issuecomment-97882692

will-moore:
@pwalczysko Thanks for testing.
It's possible that images are not resampled if they are already at a high-enough dpi. Does that explain it? Anyway, it could still be improved to explain this.
The error at https://github.com/will-moore/figure/pull/82#issuecomment-97872518 would suggest that you  are logged in as a user that doesn't have access to one of the images in the figure, which shouldn't normally happen.
I think the crop dialog still works fine with the OK being disabled until you change the selection, since until you change the selection, you have nothing to save (OK will do the same as Cancel).

will-moore:
How about this for the README:
```
 - 2_pre_resampled: This folder will only contain those images that are
   resampled in order to match the export figure resolution. This will be
   all panels for export of TIFF figures. For export of PDF Figures,
   only panels that have a 'dpi' set, which is higher than their
   existing resolution will be resampled.
```

pwalczysko:
I agree with the reformulation of the README. All is fine. Did not understand the new feature properly.

PR #87, opened 2015-06-08
Shape editor with Backbone.js

will-moore:
See https://github.com/will-moore/figure/issues/79

Adding ability to create and edit shapes on an image.

NB: Work to date has been using Backbone.js with Raphael.js shapes but this is now planned to be migrated to a new library that doesn't require Backbone.js in order that it can be used elsewhere (E.g. in OMERO webclient. 

screen shot 2015-06-08 at 10 37 38


Trefex:
Nice :)

On the topic of fill color, it could make sense to have an alpha blend instead of a "regular" fill only.

What do you think?

will-moore:
This is working so far, but will now be ported to https://github.com/will-moore/shape-editor-js to provide a shape-editor library independent of Backbone.js.

Trefex:
What was the reationale again to abandon Backbone.js?

will-moore:
I wanted to provide a shape editing tool that could be used in the webclient and other non-Backbone apps. One other example is the CRS4 OMERO team who want to add shape-drawing into a "moodle plugin". I will keep the UI (toolbar) from this PR and all the other Backbone.js code, but will move the actual shape drawing and editing to a common library to avoid code duplication etc.

Trefex:
Great, that sounds like the tool is evolving into something quite more useful than our initial request!

PR #88, opened 2015-07-01
Set up custom domain for production deployment of figure GH pages

sbesson:
This should allow to publish the figure GH pages under figure.openmicroscopy.org /cc @kennethgillen @will-moore @joshmoore @hflynn @gusferguson


will-moore:
OK, great!

PR #94, opened 2015-07-03
GitHub pages preparation for switch to master branch

sbesson:
Now that `master` and `develop` are in sync, we can switch to the default Git master branch as the development branch. If so, this PR modifies the GH pages resources to point at the commits on this branch instead of the develop one.

/cc @hflynn


hflynn:
:+1:

sbesson:
After discussion with @will-moore, merging this. `master` will be the home for stable releases and currently points at the `v1.1.1` tag while `develop` can be used for experimental work (like the `shape_editor`).

PR #100, opened 2015-08-24
Add shapes to panels

will-moore:
This is the first major part of "ROI support" in OMERO.figure.

To test:
 - Select an image in a figure
 - Choose right "Labels" tab
 - Edit ROIs button should only be enabled when a single Image panel is selected
 - Click Edit ROIs to open dialog
 - A dialog opens that allows you to draw and edit shapes on the panel
 - The 'viewport' should show the same region of the image that was shown of the figure (same zoom, pan etc)
 - Use the toolbar to pick a type of shape, line thickness & colour and then draw on image.
 - Should be able to "tweak" the last-added shape while you are adding (resize, drag etc).
 - Select shapes with the 'select' toolbar option, then resize, drag, change colour, line-width etc.
 - Copy, Paste and Delete options (under the toolbar wrench) should be enabled as appropriate and "just work"
 - Copy-then-Paste on same image should offset the new shape (so it doesn't overlap)
 - Copy, paste & delete should work with standard keyboard shortcuts.
 - Copy and Paste to a different image should be in the exact same location.
 - Once editing is done, click "OK" to close dialog.
 - Shapes should be displayed on figure.
 - Figure Undo/Redo should be updated when shapes are changed

screen shot 2015-08-25 at 21 41 22

screen shot 2015-08-25 at 21 42 23

TODO (next):
 - Multi-select shapes (drag to select)
 - Tooltip or other display of shape coordinates when dragging/resizing
 - Copy and Paste of Rectangle in "crop" dialog
 - Warning when pasting shape that is outside of viewport
 - Copy/paste, colour, thickness editing in right panel without opening dialog?

PR #101, opened 2015-09-01
Readme url fixes

will-moore:
These 2 commits were already in "develop" but not in "master" since they got left behind around the time we switch main branch from "develop" to "master".

This was noticed by @jburel because the new Shapes PR: #100 is based off "develop" and contains these commits that should really be handled separately.


sbesson:
Thanks. Merging.

PR #105, opened 2015-09-27
Shape edit 2

will-moore:
Some more copy, paste and selection functionality for ROIs.

To test:
 - Select a panel (with shapes - or add shapes to panel) and try to select multiple shapes by dragging a selection area.
 - Select All shapes via Cmd-A keyboard shortcut or via "Edit" menu.
 - The toolbar colour and line-thickness should update accordingly - E.g. if all selected shapes are same colour - that will show on toolbar. If not - will be white.
 - Updating colour or line-width in toolbar should apply to all selected shapes.
 - Copy multiple shapes to clipboard
 - Close the "Draw ROIs" dialog (with or without saving)
 - Select one or more other panels and Paste (Cmd V) to paste shapes onto the selected panels.

NB: Just noticed that I need to update the page "Edit > Paste" option to be enabled when we have copied shapes to the clipboard AND a panel is selected.


gusferguson:
@will-moore 

Tested using eel user-1 private-1.

When pasting to a smaller image only those that fit on the smaller image appear - so looks like paste failed.
When this happens in the Draw ROI window you get an error message saying shapes are outside the Viewport.
Should have same on main page as having failure without message is very confusing. I only realised there was not a paste failure because the tiniest portion of one shape appeared on the edge of the panel I pasted to.
Otherwise works as expected when pasted shapes are within the viewport.

Also noticed that when you select more than one shape and then try and move them all, only the one you click on actually  moves - all the others deselect and don’t move. I would expect all selected images to move.

will-moore:
Fixed dragging of multiple selected shapes - all should move now.
Also added some simple checking of whether pasted shapes are visible in panels they're being pasted to. NB: This is only a rough check, not very good for ellipse (only checks if centre is in viewport) but should catch most cases. User will get a message.

gusferguson:
@will-moore 

Tested using eel merge user-3 read-only-1

Behaves as expected.
Warning message good.
Multiple selected shapes all move when dragged.
Good to merge.

PR #110, opened 2015-10-14
Demo app updated with ROI features

will-moore:
This PR updates the demo app with the latest "master" OMERO.figure code.

The static demo js and html have been created from the Django-based OMERO.figure code by using a Grunt task in https://github.com/ome/figure/pull/107

To test, go to http://figure.staging.openmicroscopy.org/demo/#file/1
 - Check that the demo app is working OK
 - File > Open lists available figures that can be opened
 - ROI functionality is working same as in latest "shape edit" PR: https://github.com/ome/figure/pull/105 (can draw ROIs, copy them and paste them onto a panel).


ximenesuk:
I Open a figure, select a panel and the Draw ROIs. If I then click on Select all shaped from the Edit menu the figure disappears and I am presented with the initial New/Open figure dialog as if I had just started figure.

I can select all shapes with Cmd-A no problem, it's just the menu item that causes the problem.

will-moore:
Fixed the ROI Edit menu item bug above and fixed the "OMERO" link in top-left of app.
Also fixed the figure file http://figure.staging.openmicroscopy.org/demo/#file/1 itself:
 - Added ROIs to the figure
 - Added a Figure legend

ximenesuk:
It now works for me, the addition os some ROIs is very useful for the demo.

I do have a couple of comments.

When I choose the pointer the cursor stays as a cross (which I only expect when drawing ROIs) and only changes to an arrow when over an existing ROI. This is on Chrome 46 on MacOSX 10.10 I've not yet tried on other browsers.

When I copy a shape and then paste it to create a duplicate the new shape (or shapes) is not selected. If an existing shape is selected it stays selected. If not shape is selected then no shape is selected after the paste. I would expect the pasted shape(s) to be initially selected as the first thing a user is likely to do is reposition it/them. In addition is makes it clear which shape(s) has been pasted.

will-moore:
Thanks @ximenesuk, both issues fixed and re-deployed at http://figure.staging.openmicroscopy.org/demo/#file/1

ximenesuk:
@will-moore thanks, that was quick! Okay, one more UI feature that would be nice rather than necessary, select/deselect ROIs one-by-one using a standard key and mouse/pad. Using this I could select all, say, but then deselect one of the ROIs or select two ROIs that couldn't easily be selected by dragging around them with the pointer.

If that's too faffy or complicated then given the limited scope of this demo app I'd say this is good to merge as is.

will-moore:
@ximenesuk Thanks, I think I'll do that in a further PR. It's already on my TODO list: https://github.com/will-moore/shape-editor-js/issues/1

PR #115, opened 2015-10-28
Copy paste rois crop

will-moore:
This is the last major piece of ROIs functionality for the next release!
It allows copying and bulk editing of ROIs between panels without opening the full ROI viewer.
Also allows copying and pasting a crop region -> Rectangle and vice versa.

Functional testing is below, but any comment on workflow, usability etc would also be useful.

screen shot 2015-10-28 at 16 47 07

 - Copy & Paste ROIs
    - Draw some ROIs in the ROI dialog, OK to close.
    - The number of ROIs in selected panels should be shown at top of Labels tab.
    - Now you can copy all the ROIs on the panel with the Copy button.
    - Select another panel (or panels) on the figure and click Paste button (or keyboard shortcut)
    - Can also Delete all ROIs on selected panels with the Delete button
 - Bulk edit ROIs (color / width)
    - The Colour and Line width choosers on the ROI controls on the Labels tab should apply chosen colour or line width to all shapes in selected panels.

screen shot 2015-10-28 at 22 27 09

 - Copy & Paste crop region
    - In the Preview tab, zoom and pan the viewport, check that the View x, y, width height update.
    - Check that the "Crop" button works as before (launches crop dialog) and the "Reset" button reverts to uncropped panel.
    - Copy the crop region and paste onto another panel / panels
    - Select multiple panels with different view regions, should show '-' for x, y, width, height.

screen shot 2015-10-28 at 22 38 28

 - Copy ROIs Rectangle -> Crop region
    - Duplicate a panel by copying and pasting the panel
    - Copy a single Rectangle on the panel by selecting it and 'Copy' in the ROIs dialog.
    - Paste in the "View" controls to crop the other panel. Should show the same region highlighted by the rectangle in first panel.
 - Copy Crop region -> Rectangle
    - Again, copy and paste a panel to duplicate
    - Zoom and pan one panel to chosen view
    - Copy the view region, select the other panel and Paste this as a new Rectangle ROI.
    - Can then update the colour / line width of this as above

out0001

 - Crop dialog - pick regions from Clipboard or ROIs on the image
    - Copy some ROIs (including Rectangle(s)) to the clipboard
    - Open the Crop dialog (green button) on an image that has some Rectangular ROIs drawn on it
    - You should see that you can pick regions from the clipboard or from ROIs on the image and use these to crop the image
    - If you are cropping an image that has ROIs on it, you will be asked if you want to remove these ROIs when you crop it (since they will likely be outside or obscuring the crop region)

[...]

PR #116, opened 2015-10-29
Updating demo app with latest PR #115

will-moore:
Corresponding to #115 this ports the same changes to the demo app.


will-moore:
Since the staging demo app has been tested by @ximenesuk at https://github.com/ome/figure/pull/115#issuecomment-153385296 I think we are good to merge.

PR #118, opened 2015-11-05
Handle invalid IDs in addImages

will-moore:
This fixes https://trello.com/c/eUiSnVRV/105-404-imgdata-nan

To test, add images to a figure, entering invalid ID(s)
E.g. As plain IDs:
```
2, blah, 3
```

Or in url form:
```
http://localhost:4080/webclient/?show=image-foo|image-8102|image-bar
```

Should see a popup with a suitable message.

### DIFF
--- a/static/figure/js/views/modal_views.js
+++ b/static/figure/js/views/modal_views.js
@@ -437,10 +437,22 @@
             // The first one to return will set all the coords
             // and subsequent ones will update coords to position
             // new image panels appropriately in a grid.
+            var invalidIds = [];
             for (var i=0; i<iIds.length; i++) {
-                var imgId = iIds[i],
-                    imgDataUrl = BASE_WEBFIGURE_URL + 'imgData/' + parseInt(imgId, 10) + '/';
-                this.importImage(imgDataUrl, coords);
+                var imgId = iIds[i].replace("|", ""),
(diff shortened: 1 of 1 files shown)


ximenesuk:
This works as expected. I also noticed that if all the ids are invalid then you can't click the 'Add Images' button in the dialog.

I did enter an ID to an inaccessible image among some invalid IDs and got this:

dialogs

It was a bit odd having two distinct dialogs pop up. Is there any way these could be sequential? Is there any way they could have the same look and feel?

will-moore:
Hi @ximenesuk - I could use the plain browser's ```alert()``` for both the messages, which would make them sequential when we have both types of errors. This maybe doesn't look as nice, but trying to use the custom app warning won't work with the asynchronous calls to load all the images, without a fair bit more work. This is really supposed to be handling the odd edge case rather than expected workflow, so I'm reluctant to spend too much time on it.
Either I'd switch to using the plain ```alert()``` for everything, or I could leave it as it is?

will-moore:
@ximenesuk Now that this uses alert() for all messages, the multiple warnings you saw should all follow one after the other.

ximenesuk:
Yes, the two warnings now follow each other. Not as nice dialogs but a better workflow. Good to merge.

PR #122, opened 2015-11-10
Set line width as number, not string

will-moore:
Fixes a bug with setting line width using the controls at the top of Labels tab (which only showed as a bug when exporting the figure to PDF).
NB: I need to manually upload the export script to eel for this to be tested (export enabled).

To test:
 - Create an ROI then use the line-width drop down at top of Labels tab to change line width.
 - Try exporting as PDF.


pwalczysko:
Works as expected. please merge.

PR #124, opened 2015-11-12
Release 1.2.0 rc

will-moore:
Release candidate


will-moore:
@joshmoore Sorry - I'm not sure I understand your question.

will-moore:
@pwalczysko says "what is now on eel merge works for me" so this is good to merge...

sbesson:
The major problem raised by @joshmoore is that development versions of figure always have an exact version number encoded in their JSON file and index.html minimally. Other projects in the organization have their own post-release process to address this versioning issue: Maven projects like Bio-Formats bump their release numbers to SNAPSHOT right after release, OMERO has some internal build logic using `git describe` to version its artifacts.
We need to have a similar policy for OMERO.figure and this should likely be decided before 1.2.0 as we would like to apply it for the next development cycle.

PR #132, opened 2015-11-24
Adding 'lightweight' and updating warning in 1.2.0 blog post

will-moore:
As discussed, make clearer that ROIs are not saved to OMERO.
cc @sbesson @jburel @joshmoore

### DIFF
--- a/_posts/2015-11-24-release-1.2.0.markdown
+++ b/_posts/2015-11-24-release-1.2.0.markdown
@@ -5,7 +5,7 @@ date:   2015-11-24
 ---
 
 We are pleased to announce the release of OMERO.figure 1.2.0.
-This release introduces support for lightweight ROIs on figure panels.
+This release introduces support for illustrative ROIs on figure panels.
 
 **NB:** ROIs drawn in OMERO.figure cannot yet be saved as new ROIs on the OMERO server and it is not yet possible to load ROIs from OMERO to use in OMERO.figure.
 
--- a/_posts/2015-11-24-release-1.2.0.markdown
+++ b/_posts/2015-11-24-release-1.2.0.markdown
@@ -5,7 +5,9 @@ date:   2015-11-24
 ---
 
 We are pleased to announce the release of OMERO.figure 1.2.0.
-This release introduces support for ROIs on figure panels.
+This release introduces support for lightweight ROIs on figure panels.
+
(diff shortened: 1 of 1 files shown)


_posts/2015-11-24-release-1.2.0.markdown
```diff
@@ -5,7 +5,9 @@ date:   2015-11-24
 ---
 
 We are pleased to announce the release of OMERO.figure 1.2.0.
-This release introduces support for ROIs on figure panels.
+This release introduces support for lightweight ROIs on figure panels.
```
hflynn: I wonder if 'illustrative ROIs' would be a better term - it is actually literally true as opposed to co-opting a term which doesn't have a clear meaning in this context.

will-moore:
We just had a long discussion and 'lightweight' was the term we all agreed on. The trouble with illustrative is that OMERO doesn't distinguish between what's an illustrative ROI vv other type of ROI. All ROIs can be visualised. Even if you could save an ROI back to OMERO, it could still be "illustrative". There's not going to be one word that a user can read and understand exactly what's supported and what's not. That's why we put the full text explanation right afterwards.

will-moore:
Currently waiting on getting this merged so we can go live with the release post and continue with build jobs etc.
Do we need another call to discuss?

hflynn:
I was just trying to be helpful. I think illustrative at least has some meaning in this context whereas lightweight doesn't really, but if the consensus is in favour of lightweight, fair enough.

jburel:
We did not come up with "illustrative" during our discussion.
Final call during this afternoon meeting to see what other makes of the suggested term
then we should be a good position.

will-moore:
Apologies for the tone above. I just don't want to run out of time today - since all other release tasks are waiting on getting the blog post finalised just now.

jburel:
cc @gusferguson since the term chosen might need to be used in the help.

gusferguson:
In Help I say "Figure ROI" or "OMERO ROI" - I don't think there is any confusion in the Help.

jburel:
we decided to go for "illustrative"

will-moore:
Updated to "illustrative" and re-deployed at http://figure.staging.openmicroscopy.org/2015/11/24/release-1.2.0.html

hflynn:
good to merge

PR #136, opened 2015-12-01
Add warning of Pillow 3.0.0 bug

will-moore:
This adds a warning to the figure install instructions about a bug in Pillow 3.0.0
and how to downgrade using pip.
Staged at http://figure.staging.openmicroscopy.org/#install

### DIFF
--- a/index.html
+++ b/index.html
@@ -166,6 +166,20 @@ title: OMERO.figure
                 $ path/to/bin/omero script upload omero/figure_scripts/Figure_To_Pdf.py --official
               </code>
             </li>
+            <li><strong>Warning: Pillow 3.0.0 bug breaks figure export. </strong><br>
+              PIL (or Pillow) is required for exporting figures and should already be installed as an OMERO.web
+              dependency.
+              We have found that a
+              <a href="https://github.com/python-pillow/Pillow/issues/1524">bug in Pillow 3.0.0</a>
+              causes OMERO.figure export to fail.
+              If you have Pillow 3.0.0 you can fix this by downgrading to Pillow 2.9.0.
+              For example, if you have installed with pip you can downgrade with:
+                <code style="width:700px"><pre>
+$ python -c "import PIL; print PIL.PILLOW_VERSION"
+3.0.0
+$ pip uninstall pillow
+$ pip install pillow==2.9.0</pre></code>
+            </li>
           </ul>
         </p>
 


aleksandra-tarkowska:
Why just not add requirement file?

manics:
It might be installed as a distribution package (e.g. rpm, deb) instead of by pip, in which case you might be running a different version.

hflynn:
I can open a PR to add this the the OMERO docs assuming the same effect on other scripts. You've confirmed 2.9.0 works and this just affects 3.0 right?

ximenesuk:
As far as the PR goes the advice works. If this is the way to go, given the discussion above, I'd suggest splitting the code since the first line is about checking the installed version rather than downgrading. Something more like:



If you have Pillow 3.0.0 you can fix this by downgrading to Pillow 2.9.0.

Check the installed version of Pillow:
```
$ python -c "import PIL; print PIL.PILLOW_VERSION"
3.0.0
```
If you do have Pillow 3.0.0 then, for example, if you have installed with pip you can downgrade with:
```
$ pip uninstall pillow
$ pip install pillow==2.9.0
```

will-moore:
@aleksandra-tarkowska I don't know if this update is enough to justify adding a requirements file. I've no experience of requirements files so it's not something I've looked into. But that may be a useful idea, especially if the number of requirements increases.

@manics Certainly if you are running a different version you won't need to do anything. If you used a different package manager then you'll need different instructions for downgrading. I didn't want to go into all that detail here, and merely provided the ```pip install``` instructions since that's what I know.
I don't have an idea how many of our users use pip vv other install methods. If anything, I'd prefer to remove the pip instructions and let all admins figure it out, rather than provide instructions for all different scenarios.

@hflynn I think it would be useful for this to be included in the appropriate place in OMERO docs. Also to change the version in the omero requirements file. I've confirmed that 2.9.0 works and I think it only affects 3.0.0 (and future releases until it's fixed I guess).

@ximenesuk I think your suggestion makes sense, I'll update...

manics:
@will-moore My comment was against the addition of a requirements file. I think the example in your docs is fine.

joshmoore:
> It might be installed as a distribution package (e.g. rpm, deb) instead of by pip, in which case you might be running a different version. ...  I think the example in your docs is fine.

I'd disagree. I think we it should be an "AND" rather than an "OR". That gives us:

 * docs
 * requirements.txt
 * setup.py specification (this can be read from requirements.txt if necessary)

It's redundant, but this is currently the "state-of-python" at the moment. I'd like to see us work towards having requirements.txt everywhere with the goal of eventually having whatever package manager (virtualenv, conda, etc) be able to detect conflicts.

will-moore:
OK, requirements.txt seems pretty easy. Can you do "If python 2.6, you need reportlab 2.7" as we have in the docs?
Also, if you've already got Pillow 3.0.0, and you have ```pillow==2.9.0``` in your requirements.txt, does that replace the existing Pillow?
Currently the figure install docs don't mention Pillow at all, since it's already required by OMERO.web.
Really, this would be best moved to [python_deps.sh](https://github.com/openmicroscopy/openmicroscopy/blob/develop/docs/install/python_deps.sh) and https://www.openmicroscopy.org/site/support/omero5.2/sysadmins/unix/server-installation.html#python-2 and https://www.openmicroscopy.org/site/support/omero5.2/sysadmins/unix/server-linux-walkthrough.html and https://www.openmicroscopy.org/site/support/omero5.2/sysadmins/unix/server-centos6.html and all the install scripts.

hflynn:
Changes are already underway on the OMERO installation pages but I'll make sure the Pillow issue gets added before we are finished.

manics:
> OK, requirements.txt seems pretty easy. Can you do "If python 2.6, you need reportlab 2.7" as we have in the docs?

I think you'd need seperate requirements files for Python 2.6 and 2.7

> Also, if you've already got Pillow 3.0.0, and you have pillow==2.9.0 in your requirements.txt, does that replace the existing Pillow?

Not sure, though it should at least give you a warning. Does it have to be 2.9.0, or will any 2.X do (e.g. centos-7 has pillow 2.0.0)? Would something like `2<=pillow<3` work?

will-moore:
I was hoping to quickly check if the pillow bug is actually a problem on centos (or anything other than my mac) but I've spent all morning trying to get OMERO running on docker and failed. So if someone else wants to test that, it would be great.

will-moore:
Running scripts from CLI in docker container seems to suggest that Pillow 3.0.0 is working for Batch Image Export of TIFFs

```
Callback received: FINISHED

	*** start stdout (id=43)***
	* {'Export_Individual_Channels': True, 'Choose_Z_Section': 'Default-Z (last-viewed)', 'Choose_T_Section': 'Default-T (last-viewed)', 'Data_Type': 'Image', 'Format': 'TIFF', 'IDs': [1L], 'Zoom': '100%', 'Export_Merged_Image': True, 'Folder_Name': 'Batch_Image_Export', 'Individual_Channels_Grey': False}
	* Processing 1 images
	* Exporting image as TIFF: test.fake
	* 
	* ----------- Saving planes from image: 'test.fake' ------------
	* Using:
	*   Z-index: Last-viewed
	*   T-index: Last-viewed
	*   Format: TIFF
	*   Image Zoom: 100%
	*   Greyscale: False
	* Channel Rendering Settings:
	*   0: 0-255
	* 
	* savePlane..
	* channel: merged
	* z: 1
	* t: 1
	* Saving image: Batch_Image_Export/test.fake_merged_z01_t01.tiff
	* 
	* savePlane..
	* channel: c00
	* z: 1
	* t: 1
	* Saving image: Batch_Image_Export/test.fake_c00_z01_t01.tiff
	* Duration: 0:00:00.790522
	* 
	*** end stdout ***
```

Just need to check what's actually in the file annotation...

will-moore:
Just been checking why Batch_Image_Export can save PIL to tiff but figure can't: It's because figure is converting images to RGBA (which allows white background when image is rotated). Even though they are being converted back to RGB, this is not enough to fix the issue.

will-moore:
Since we now have OMERO installs using Pillow < 3.0.0 I think we don't need to duplicate this requirement in OMERO.figure.

PR #138, opened 2015-12-14
Remove bin/omero web stop & start from figure docs

will-moore:
As suggested by @jburel, not all users will restart web with ```$ bin/omero web stop``` and ```$ bin/omero web start```. Instead, simply instruct users to restart web according to their deployment setup.


ximenesuk:
Looks fine to merge.

PR #142, opened 2016-02-18
imgData pixel sizes in custom units, not MICROMETERS

will-moore:
This fixes a bug in OMERO.figure caused by the fix in https://github.com/openmicroscopy/openmicroscopy/pull/4460 and discussed in https://github.com/openmicroscopy/openmicroscopy/pull/4460#issuecomment-181951303.

Since the imgData json from webgateway now alway uses MICROMETERS for pixel size units, when the units are different OMERO.figure needs to use the value for the custom unit.

To test, set the pixel size for an image to use different units as described https://www.openmicroscopy.org/site/support/omero5.2/developers/Python.html#create-image.
Add that image to a figure and see that the pixel size shown in right panel is correct.

### DIFF
--- a/views.py
+++ b/views.py
@@ -136,9 +136,14 @@ def imgData_json(request, imageId, conn=None, **kwargs):
     rv = imageMarshal(image, request=request)
 
     if unitsSupport:
+        px = image.getPrimaryPixels().getPhysicalSizeX()
+        # overwrite existing x value, which will be in MICROMETERS
+        rv['pixel_size']['x'] = px.getValue()
         rv['pixel_size']['symbolX'] = px.getSymbol()
         rv['pixel_size']['unitX'] = str(px.getUnit())
         py = image.getPrimaryPixels().getPhysicalSizeY()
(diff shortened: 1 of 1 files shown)


pwalczysko:
Did some prep of data, will finish on Monday.

chris-allan:
Technically I don't see a problem with this approach. It feels a bit ugly to be replacing those properties like that when they are semantically different in another context with effectively the same naming but it certainly works.

pwalczysko:
Tested with real files with units of
   - dm
   - mm
   - nm

All worked fine as checked per display in Web and Insight comparing with
screen shot 2016-02-22 at 18 14 30

will-moore:
@chris-allan Agreed - that was just the quickest fix possible! But I'll look into adding different attributes for the non-converted values, and leave ```x``` and ```y``` in Microns....

will-moore:
This has been re-opened as https://github.com/ome/figure/pull/144 with new branch and rebased on latest master.

PR #145, opened 2016-03-15
File open dialog

will-moore:
This improves the File > Open dialog in a couple of ways.
Faster loading of files from OMERO (using projection query and not parsing json).
We now parse the json from file description during the Javascript render() of files.
The handling of missing thumbnails for deleted images is now done by the webgateway render_thumbnail (same as for webclient) instead of doing an extra query to identify missing images.

To test:
 - Try opening files in figure. Check for any problems. Should work as before.
 - When files are being loaded (and on "Refresh"), you should see a "Loading Files..." placeholder.
 - Loading of files should be faster, although I've previously only noticed it being slow on nightshade (7 secs to load 135 files).
 - The currently opened file should be "disabled" (not clickable) in the File > Open list. Should also be "disabled" after hitting refresh.
 - Try opening another user's File (where canEdit is False) and check that this is still reflected in the UI - shouldn't be able to Save change.


pwalczysko:
All works as expected for every test. Very hard to compare the speed, it seems fast though on eel merge. But even on eel latest the loading is pretty fast. 
Ready to merge.

will-moore:
To test the last commit:
 - save a figure file named with some script html in it, E.g.  ```test<script>alert('bad')</script>```.
 - shouldn't get any alert popups - name should be displayed correctly
 - Then go to File > Open and check that the name is fully rendered in the file list and no alert popups are shown.
 - Also try refreshing the page or starting from the ```/figure/``` home page and going to File > Open.

screen shot 2016-04-07 at 12 09 18

pwalczysko:
Works as expected. Ready to merge.

will-moore:
@joshmoore / @jburel Would be great to get this merged if possible, then I can fix conflicts with https://github.com/ome/figure/issues/147 - Thanks!

will-moore:
@joshmoore / @jburel / @sbesson If you could merge this and https://github.com/ome/figure/pull/147 then I can prepare release RC1? Thanks.

PR #152, opened 2016-05-04
open with images

will-moore:
This adds support for directly creating a new figure with chosen images by going to
```
 /figure/new/?image=1&image=2
```

To test, use 1 or more image IDs to construct a url in the form above.
Navigate to this url and this should display a new figure with the images already added.
Then you can edit etc as normal.
- Try opening a new file without saving (expected warning and behaviour?)
- Back button will re-create figure with images as before.

Also need to test manually adding images using image IDs or link url copied from webclient or 'external' image viewer url, E.g. http://jcb-dataviewer.rupress.org/jcb/img_detail/635357/29155


gusferguson:
@will-moore 

Tested with https://cowfish.openmicroscopy.org/merge/webclient/ user-3 read-only-1

Warning is as expected.
Behaves as expected.
Good to merge.

will-moore:
@jburel Can we merge this so I can fix conflicts with #160?

PR #161, opened 2016-08-04
Disable save button

dominikl:
Simply disables the 'Save Figure' button right after the user clicks on it, to prevent multiple simultaneous save actions being triggered. See [Trello - Disable 'Save' when clicked](https://trello.com/c/mZsyPq8K/130-disable-save-when-clicked)


will-moore:
Looks good and works fine.
Good to merge.

PR #162, opened 2016-08-04
Import from json

dominikl:
Adds the possibility to create a Figure from a (previously exported) JSON object, see [Trello - import Figure from json](https://trello.com/c/Ad1cd1fA/73-import-figure-from-json).

I'm not sure, if that's the correct way to add stuff to models, views, etc. please correct me, if not. Especially listening on the form submit within `figure_view.js` seems to be a bit out of place.


will-moore:
This looks great and works nicely.

I wonder if you could avoid the code duplication by calling the new ```figure_fromJSON(data)``` from within ```load_from_OMERO()``` above. You might need to add fileId to the ```data``` object, but otherwise the code looks pretty similar. You can use the ```self``` handle in the getJSON callback, E.g. ```self.figure_fromJSON(data)```.

Cheers!

static/figure/js/models/figure_model.js
```diff
@@ -45,47 +45,53 @@
...
+              data.fileId = fileId;
+              self.load_from_JSON(data);
             });
+            
+            self.set('unsaved', false);
         },
-
+        
+        load_from_JSON: function(data) {
+          self = this;
```
will-moore: This needs a ```var```:
```var self = this;```
Also it would be great to keep all indents at 4 spaces so we are consistent everywhere.

File: static/figure/js/models/figure_model.js [hunk 2]
```diff
@@ -45,47 +45,53 @@
...
-                self.set('unsaved', false);
-                // wait for undo/redo to handle above, then...
-                setTimeout(function() {
-                    self.trigger("reset_undo_redo");
-                }, 50);
+              data.fileId = fileId;
+              self.load_from_JSON(data);
             });
+            
+            self.set('unsaved', false);
```
will-moore: This line needs to be called within the ```$.getJSON`` callback above, right after ``self.load_from_JSON(data)```. Currently this is getting called before the data loads.

will-moore:
Thanks! Looks great. Tested locally.
Good to merge.

static/figure/js/models/figure_model.js
```diff
@@ -104,7 +110,7 @@
 
             return json;
         },
-
+
```
will-moore: @dominikl I wonder if you could remove this whitespace change (and see if there are any others that could be cleaned up)?
I don't know if you're using ```$ git add -p``` but I find this can help avoiding unnecessary whitespace changes?
Thanks

PR #169, opened 2016-08-30
Xywh panel 

bramalingam:
Testing instructions : 

1) Try manually manipulating X,Y Width and Height values in the info panel (both single and multi-select cases).
Check if the changes get updated on the panels.
2) Try dragging an image from one page to another and check if the X,Y values are reset to 0,0 at the start of every new page.
3) Multi-select : Select 2 images from 2 different pages (1 image from page 1 and 1 image from page 2) and change X,Y values. Check if the images get aligned to the appropriate X,Y positions within the page that they reside in. (i.e : image 1 is set to X,Y in page 1 , and image 2 is set to the same X,Y in page 2)

@will-moore


static/figure/js/views/right_panel_view.js
```diff
@@ -692,6 +692,33 @@
...
+                if (attr === 'x' || attr ==='y'){
+                    var old = m.get(attr);
+                    var coords = {};
+                    coords[attr] = old;
+                    var offset = this.figureModel.getPageOffset(coords);
+                    m.set(attr, old - offset[attr] + value);
+                }
+                else {
+                    console.log(attr);
+                    console.log(value);
```
dominikl: The `console.log`s  are probably not necessary, are they?

dominikl:
Looks good from my point of view... Functionality works like expected.

will-moore:
Only the ```x``` input is ```type='number'```. I noticed this because Chrome won't set '-' for a number and you see an error in console:

screen shot 2016-08-31 at 17 11 59

I think x, y, width & height need to be "number" inputs.
It might be nice to also use an empty string instead of "-" for x,y,w,h (but not for the other attrs) so you don't force the browser to catch the error.

static/figure/figure.js
```diff
@@ -5338,17 +5374,24 @@ var RectView = Backbone.View.extend({
                     remoteUrl = m.get('baseUrl') + "/img_detail/" + m.get('imageId') + "/";
                 }
                 // start with json data from first Panel
+                var this_json = m.toJSON();
+                // Format floating point values
+                _.each(["x", "y", "width", "height"], function(a){
+                    if (this_json[a] != "-") {
```
will-moore: I noticed you can remove this ```if``` check now, since ```m.toJSON()``` won't give you any ```-``` values.

will-moore:
Need to fix the setting of minus values for x & y when not on page 1 (seems to always jump to the previous page).
Also need to disallow width and height to be less than 1 (negative is bad, probably zero is bad too).
You can set a ```min='1'``` when you make these into 'number' inputs, but also have to handle this in JavaScript since users can still type in negative numbers.

will-moore:
Working nicely now. No console errors, re-render bug fixed.
Good to merge.

PR #176, opened 2016-10-11
Fix highlighter to gh-pages supported one

hflynn:
Although I fixed the jekyll build as my local installation instructed, it seems gh-pages does not support pygments and we got an error for the website build when the live site was updated. This should fix it without altering the appearance of the site, as it's apparently what gh-pages is using now anyway.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -1,5 +1,5 @@
 name: Your New Jekyll Site
 markdown: kramdown
-highlighter: pygments
+highlighter: rouge
 
 baseurl: /figure


hflynn:
For reference:
"The page build completed successfully, but returned the following warning:

You are attempting to use the 'pygments' highlighter, which is currently unsupported on GitHub Pages. Your site will use 'rouge' for highlighting instead. To suppress this warning, change the 'highlighter' value to 'rouge' in your '_config.yml' and ensure the 'pygments' key is unset. For more information, see https://help.github.com/articles/page-build-failed-config-file-error/#fixing-highlighting-errors.

For information on troubleshooting Jekyll see:

  https://help.github.com/articles/troubleshooting-jekyll-builds"

hflynn:
Staged at http://figure.staging.openmicroscopy.org and build is green https://ci.openmicroscopy.org/view/Docs/job/FIGURE-help-staging/52/ so you just need to check staging looks the same as the live site really

jburel:
I did not notice any difference between staging and live versions

PR #178, opened 2016-10-12
code layout

will-moore:
This moves all the source code js and templates files out of the Django app static directory
into a new top-level `src` directory.

The grunt `jst` and `concat` tasks compile templates and concatenate js files into `omero_figure/static/figure/templates.js` and `omero_figure/static/figure/figure.js`.

All subsequent PRs will need to be on top of this one to avoid conflicts.
To test:
- `$ grunt watch`
- Edit a js file under `src/js`, Save -> Grunt should concat -> figure.js
- Edit a template under `src/templates`, Save -> Grunt will compilte -> templates.js

NB: removed the build date that is updated with every concat of figure.js to reduce conflicts between PRs.

cc @bramalingam Best test of this PR is to begin a new feature on top of it.

PR #187, opened 2016-11-30
Maintain aspect ratio Button

bramalingam:
This PR adds an extra button to control the aspect ratio of the panels, when the user edits the width/height of the image.

To Test:
1) Test if toggling the button changes the colour to red(inactive) and green(active) states.
2) When the button is red, edit width or height of the image. This will just change the width of height of the image.
3) When the button is green, edit width or height of the image. Please check the width/height ratio (aspect ratio) of the panel before editing. Post editing the width/height of the panel, the height/width of the image should automatically get calculated and adjusted. The resultant panel aspect ratio, should be exactly the same as the previous aspect ratio.
4) Check if the tooltip on the button makes sense.

And as always code review would be useful as well.
The branch includes the code-layout branch as well, and the PR could technically be considered as an independent review of the code-layout PR from @will-moore 
@will-moore @jburel


Gruntfile.js
```diff
@@ -8,32 +8,32 @@ module.exports = function (grunt) {
     , jshint: {
         all: [
             "Gruntfile.js"
-          , "omero_figure/static/figure/js/models/*.js"
-          , "omero_figure/static/figure/js/views/*.js"
-          , "omero_figure/static/figure/js/app.js"
+          , "src/js/models/*.js"
```
aleksandra-tarkowska: you have to rebase this branch as is outdated

will-moore:
This PR is on top of Code layout changes in https://github.com/ome/omero-figure/pull/178 which should be merged first. Then this PR can be rebased (or merge master back into this PR).

will-moore:
You shouldn't need to add an ```<input>``` element to the div to get the connecting line. The ```<input>``` can be focussed, which doesn't look great:
 
screen shot 2016-12-01 at 12 21 05

The 'link' button could do with a tool-tip (add a "title" attribute).
Also, I think the red & green colours are too aggressive.
Red should really only be for errors or delete buttons. Maybe best to stick with the default colour of the button for 'off'. Maybe green is OK for 'checked' but I wonder if there's a nicer solution. It's not that often you see green to mean 'checked'.

For some reason the ```<span>``` in the link button doesn't have ```cursor:pointer``` so you get a regular arrow cursor here although you get the expected pointer cursor on the surrounding button.

Minor bug - when I first click on the 'width', the value says '100' but I think the value in the model is actually a floating point value E.g. '100.1' because when it loses focus (without me changing the value) I get a re-render (link button 'unchecked') because it is setting width to new value of '100'.
When I repeat this I don't see the same bug because now the value is actually '100' in the model.
Seems that the re-checking of link button after render() is not working?

Also noticed an existing tiny bug: When I set width or height, the "Save" button at the top of the screen doesn't get enabled. To enable it, you need to use ```m.save(newValues)``` instead of ```m.set(newValues)``` because I'm using different underlying Backbone events to trigger the "Save" button and these events come from model.save(). This really needs to be cleaned-up in due course, but for now, just use m.save().

will-moore:
Just tried using a sibling ```<span class="glyphicon glyphicon-ok-sign"></span>``` with some extra styling (see below) to show a green check. So, something like that might work? You'd just need to only show it when the parent button has the 'active' class, which I can show you how to do.

screen shot 2016-12-01 at 13 02 48

<details><summary>span styles...</summary>

```
    color: green;
    position: relative;
    left: -22px;
    top: -7px;
    font-size: 10px;
```
</details>

jburel:
@bramalingam could you sort the travis failure? so it can be included in the daily build

PR #191, opened 2017-01-12
License expression

jburel:
Use valid SPDX license expression

To test: 
 * run ``python setup.py sdist``
 * Check that there is no message like ``npm WARN omero-figure@2.0.1 license should be a valid SPDX license expression
`` 
 see  https://spdx.org/licenses/


will-moore:
```
$ python setup.py sdist
...
Creating tar archive
removing 'omero-figure-2.0.1' (and everything under it)
```
Didn't see any npm license warnings.
Good to merge.

will-moore:
Good to merge.

PR #193, opened 2017-01-24
Xywh width height button

bramalingam:
This PR adds an extra button to control the aspect ratio of the panels, when the user edits the width/height of the image.

To Test:

Test if toggling the glyphicon-link button changes status of glyphicon-ok-sign to show/hide.
When the glyphicon-ok-sign is hidden, edit width or height of the image. This will just change the width or height of the image (alone).
When the glyphicon-ok-sign is shown, edit width or height of the image. Please check the width/height ratio (aspect ratio) of the panel before editing. Post editing the aspect ratio should be maintained. 
Check if the tooltip on the button makes sense.
And as always code review would be useful as well.
The branch includes the code-layout branch as well, and the PR could technically be considered as an independent review of the code-layout PR(#178) from @will-moore
@will-moore @jburel @aleksandra-tarkowska 

Conversation trace from closed PR: #187


will-moore:
Looks good and is working well.
Minor bug: when the panel is first rendered, you have the green 'check' showing even though the link button is not 'checked'.
You could show/hide the green check purely using the "active" class on the parent button.
```
  .active .aspect_ratio_selected { display: inline-block }
  .aspect_ratio_selected {
    display: none
    }
```
Actually it would also be nicer to add the other styles 'color: green' etc to the css .aspect_ratio_selected class instead of in-line styles.

screen shot 2017-01-24 at 12 12 58

Also, as mentioned before, you need to use ```m.save(newValues)``` instead of ```m.set(newValues)``` to enable the figure "Save" button.

will-moore:
You need to remove the compiled figure.js and templates.js files. I thought the 'merge' had removed them but it seems not.

src/js/views/right_panel_view.js
```diff
@@ -747,14 +749,43 @@
...
                 }
             }.bind(this));
             // Timout for ignoreChange
             // Only reset this AFTER render() is called
             setTimeout(function(){
                 this.ignoreChange = false;
-            }, 50);
+                // keep locked status of the aspect ratio button the same,
+                // when the focus shifts because of a blur event
+                if (aspectRatioStatus === true) {
```
will-moore: Don't need ``` === true``` here. Just ```if (aspectRatioStatus) ```

src/js/views/right_panel_view.js
```diff
@@ -777,6 +808,17 @@
...
 
+        lockAspectRatio: function(event) {
+            event.preventDefault();
+            $(".set_aspect_ratio", this.$el).toggleClass("active");
+            if ($(".set_aspect_ratio", this.$el).hasClass("active")) {
+                $(".aspect_ratio_selected", this.$el).addClass("glyphicon-ok-sign");
+            }
+            if (!$(".set_aspect_ratio", this.$el).hasClass("active")) {
+                $(".aspect_ratio_selected", this.$el).removeClass("glyphicon-ok-sign");
+            }
```
will-moore: If you style the ```. aspect_ratio_selected``` to only be visible when the parent is ```.active``` (see other comment) then you can get rid of all this extra logic.

src/templates/xywh_panel_template.html
```diff
@@ -33,18 +33,27 @@
                         <label class="col-sm-2 control-label">Width</label>
                         <div class="col-sm-4">
                             <input type="number" name="width" min="0" class="form-control input-sm" value="<%= width %>">
-                        </div>
-                    </div>
-                    <div class="form-group" style="margin-bottom: -15px;">
+                        </div>                     
+                    </div>
```
will-moore: The only diff here is adding whitespace. Would be nice to clean that up (as you mentioned).

bramalingam:
@will-moore I have pushed two commits. Please check and let me know if it looks fine.

will-moore:
@bramalingam If I check the 'link', change the Width then move focus away (E.g. click on X input) then the panel is re-rendered, the 'link' button is maintained as '.active' but we've lost the green check.
See my previous comments on how to tie the 'green check' to the .active class via css instead of JavaScript, so you don't need to update them both at the same time in JS.

Also, as mentioned before (twice), you need to use m.save(newValues) instead of m.set(newValues) to enable the figure "Save" button.

jburel:
@bramalingam: ready for review?

bramalingam:
Yes it is, guess @will-moore has already reviewed the changes but yeah I can re-list on for review tomorrow.

jburel:
@will-moore could you check the last commits?

src/templates/xywh_panel_template.html
```diff
@@ -47,12 +47,8 @@
...
                         </div>
                     </div>
-                    <button type="button" class="btn btn-link set_aspect_ratio"
-                        style="position:absolute; left:195px; top:25px; padding:3px; outline:none !important";
-                        title="Maintain aspect ratio">
-                            <span class="glyphicon glyphicon-link"></span>
-                            <span class="glyphicon glyphicon-ok-sign aspect_ratio_selected" style="left:187px; top:-51px; font-size:10px; color:green"></span>
-                    </button>
+                    <button type="button" class="btn btn-link setAspectRatio"
+                        style="position:absolute; left:195px; top:25px; padding:3px; width:23px; height:24px; outline:none !important";title="Maintain aspect ratio"><span class="glyphicon glyphicon-link"></span><span class="glyphicon glyphicon-ok-sign okSign" style="left:-28px; top:-7px; font-size:10px; color:green"></span></button>
```
will-moore: The button tooltip isn't working because you've got ``` ;title ``` there (remove the ;)

will-moore:
Working nicely now - Just a tiny typo fix for ```;title```.
Otherwise good to merge.

jburel:
@will-moore happy now with the last commit

will-moore:
Yes, looks good now thanks.

PR #197, opened 2017-03-01
Introduce dev version

jburel:
Add dev version 
e.g. https://pypi.python.org/pypi/estimators/0.1.0.dev0

cc @sbesson

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "omero-figure",
-  "version": "2.0.1",
+  "version": "2.0.1.dev0",
   "description": "OMERO figure creation app",
   "main": "index.js",
   "scripts": {
--- a/setup.py
+++ b/setup.py
@@ -42,7 +42,7 @@ def read_file(fname, content_type=None):
     return data
 
 
-VERSION = "2.0.1"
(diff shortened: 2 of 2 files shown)


joshmoore:
In http://semver.org/ think this would be `2.0.1-dev0`, but see also https://ziade.org/2011/02/15/do-not-upload-dev-releases-at-pypi/

Other options for these SNAPSHOT-like builds of Python resources:
 * PEP-503 (https://pypi.python.org/pypi/pypiserver/, https://pypi.python.org/pypi/devpi-server,  etc)
 * Artifactory
 * Conda

jburel:
I don't think we will push the dev version to pypi. The problem is mainly  with testPypi, used  a staging mechanism to test the full workflow
we have to bump the version each e.g. figure has a higher number (I was testing the build)
so it is "easy" to forget to roll back the version when we are ready

joshmoore:
> I don't think we will ever push the dev version to pypi. 

:+1:

> The problem is mainly with testPypi

Understood, but I think some of the other suggestions could potentially help there. But we'll need time to investigate.

jburel:
@joshmoore: Agree This is just to see an "option in action"

PR #198, opened 2017-03-24
Fix label delete

will-moore:
Fixes the deletion of labels that contain unescaped characters, e.g. single quotation marks ``` ' ```.

To test, create, labels with various escaped characters in ``` < > ' " ```.
Check you can update these labels (change text or color etc) and delete them.

### DIFF
--- a/src/js/views/right_panel_view.js
+++ b/src/js/views/right_panel_view.js
@@ -430,6 +430,8 @@
                 key = $form.attr('data-key'),
                 deleteMap = {};
 
+            // escape the key to handle 'single' and "double" quotes
+            key = _.escape(key);
             deleteMap[key] = false;
 
             this.models.forEach(function(m){


waxenegger:
I think a combination of " and ' it doesn't like: http://web-dev-merge.openmicroscopy.org/figure/file/18701

The combination of same text labels (and treating them as one for the purpose of edit/deletion) is a feature, right?

will-moore:
@waxenegger Yes, identical labels are grouped for editing (for editing the same label on lots of selected panels at once).

will-moore:
I removed the last commit 0b4918c which reverted 68610373262db623 since I didn't want to revert a revert-commit!
Fixed the bug by also escaping the key used in delete (now we escape this for both edit and delete so they are both working with single and double quotes).

waxenegger:
Label delete is working fine now, for single and double quotes as well.

jburel:
Discussed with @will-moore, we may want to review at some points the addition of the same label when we only have one image with one channel!
Merging the fix

PR #200, opened 2017-03-31
Slider bugfixes

will-moore:
Fix the display of Timestamp and LUT background for channel sliders.

Previously, timestamp label (for movies with timestamp info) was overlapped by the channel sliders and the channel sliders weren't showing LUT background correctly.

See before and after shots:

screen shot 2017-03-31 at 15 17 25

screen shot 2017-03-31 at 15 16 42

To test:
 - Find an image with timestamp info. Check that label is not obscured.
 - Test display of LUTs on channel sliders.

### DIFF
--- a/omero_figure/static/figure/css/figure.css
+++ b/omero_figure/static/figure/css/figure.css
@@ -711,7 +711,7 @@
         display: inline-block;
     }
 
-    .lutBg {
+    .lutBg, .ch_slider .ui-slider-range {
         /* NB: when updating png, consider using different name to avoid cache */
         background-image: url('../images/luts_10.png');   /* each lut is 10px high */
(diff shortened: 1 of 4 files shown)


waxenegger:
Works nicely, both space for timestamp and luts background for slider.

figure_timestamp

omero_figure/static/figure/css/figure.css
```diff
@@ -711,7 +711,7 @@
         display: inline-block;
     }
 
-    .lutBg {
+    .lutBg, .ch_slider .ui-slider-range {
         /* NB: when updating png, consider using different name to avoid cache */
         background-image: url('../images/luts_10.png');   /* each lut is 10px high */
```
jburel: own copy
this is the "out of synch" I am referring to in https://github.com/openmicroscopy/openmicroscopy/pull/5210

PR #205, opened 2017-04-18
Update readme

jburel:
requirement bump to 5.3.0
cc @will-moore 

Noticed while checking https://testpypi.python.org/pypi?:action=display&name=omero-figure&version=3.0.0


will-moore:
Makes sense. Good to merge.

PR #218, opened 2017-06-02
basic working flip behaviour

will-moore:
See https://trello.com/c/wlX18bfE/128-flip-vertical-or-horizontal

Work in progress.

cc @carandraug

TODO:

 - [ ] Handle drag/pan in right view-port with  ```right_panel_view.js``` ```correct_rotation(dx, dy, rotation)```
 - [ ] Maybe ```flip``` variable on PanelModel should be ```flip_x``` in case we decide in this PR or later that we also need a ```flip_y``` option too?
 - [ ] We probably need to move the Channel toggle buttons down, alongside the channel sliders, in order to make more space for a Flip button (should be a different PR). Flip button could be a single toggle button (like the Z-projection button) or if we want flip horizontal and vertical then a button might have a drop-down with checkbox menu for Vertical and Horizontal.
 - [ ] In the ROI dialog, we need to cancel the flip (same as we do with rotation) so that drawing isn't broken with the same drag issues we see in the viewport.
 - [ ] Export script needs to handle flipped images for PDF and TIFF export

### DIFF
--- a/src/js/models/panel_model.js
+++ b/src/js/models/panel_model.js
@@ -18,6 +18,7 @@
             selected: false,
             pixel_size_x_symbol: '\xB5m',     // microns by default
             pixel_size_x_unit: 'MICROMETER',
+            flip: true,
 
             // 'export_dpi' optional value to resample panel on export
             // model includes 'scalebar' object, e.g:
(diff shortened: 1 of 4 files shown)


will-moore:
Closing for now, until more complete...

PR #220, opened 2017-06-14
Update shape-editor-3.0.0

will-moore:
This updates shape-editor.js to the latest version (not tagged yet) which will be 3.0.0.
See https://github.com/ome/omero-figure/pull/213.

To test latest fix, check that "strange borders" are not seen when zoomed in a lot to tiny images as shown in https://github.com/ome/omero-figure/pull/213#issuecomment-300137534
(This fix was previously tested in that PR but reverted there to go into an updated shape-editor.js).


jburel:
I will need to tag shape-editor

pwalczysko:
Bug: the whole Figure seems to be not working now. No white canvas appears, File > Open does not work.

``ReferenceError: ShapeManager is not defined``

jburel:
everything is back to normal

PR #228, opened 2017-06-28
Export figure without script

will-moore:
See https://trello.com/c/FMMoy6PD/6-export-without-scripting-service

Work in progress. TODO:

 - [ ] Move export script into sensible location (not scripts dir)
 - [ ] Support export of zip (including multi-page TIFFs)
 - [ ] Handle export to OMERO images. AJAX instead of form submit, show link to Image (same as now)
 - [ ] Move Fonts into figure repo

Future PRs:
 - Allow to export from figure.json file ID
 - CLI option for exporting (omero plugin?)


omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
```diff
@@ -1404,12 +1415,14 @@ def paste_image(self, pil_img, img_name, panel, page, dpi):
         # in the folder to zip
         if self.zip_folder_name is not None:
             img_name = os.path.join(self.zip_folder_name, FINAL_DIR, img_name)
+            pil_img.save(img_name)
 
-        # Save Image to file, then bring into PDF
-        pil_img.save(img_name)
+        buffer = StringIO()
```
jburel: will be better to start using ``io.StringIO`` if we want to start moving to python3

requirements.txt
```diff
@@ -1 +1,3 @@
-omego
\ No newline at end of file
+omego
+reportlab
```
jburel: we may have some surprise as we had with pillow if we install from pip
That's why we removed them from the installation doc

will-moore:
@jburel OK, maybe I'll try and install on devspace first.
Excluded by travis failures for tomorrow anyway, so won't break the build.

PR #229, opened 2017-06-29
reverseIntensity keyError export

will-moore:
Fixes bug with export of images that were added to figures before 'reverseIntensity' was supported (before 5.3.0).
Reported to @kennethgillen from nightshade:
https://trello.com/c/baovPrAJ/7-reverseintensity-keyerror-in-export

To test:
 - I reproduced the bug by manually removing 'reverseIntensity' info from the JSON:
     - File -> Export as JSON
     - Copy this data, removing the ```, "reverseIntensity": true``` (only need to do this for a single active channel of a single image.
     - Create a new figure with this text via File -> Import from JSON.
 - Export of this figure should not cause any error.

cc @jburel - We should probably release this as a 3.1.1 bug-fix ASAP.

### DIFF
--- a/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
+++ b/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
@@ -759,7 +759,7 @@ class FigureExport(object):
                 c_idxs.append(i+1)
                 windows.append([c['window']['start'], c['window']['end']])
                 colors.append(c['color'])
-                reverses.append(c['reverseIntensity'])
+                reverses.append(c.get('reverseIntensity', False))
 
         image.setActiveChannels(c_idxs, windows, colors, reverses)
 


omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
```diff
@@ -759,7 +759,7 @@ def apply_rdefs(self, image, channels):
                 c_idxs.append(i+1)
                 windows.append([c['window']['start'], c['window']['end']])
                 colors.append(c['color'])
-                reverses.append(c['reverseIntensity'])
+                reverses.append(c.get('reverseIntensity', False))
```
sbesson: Generally  :+1: for generalizing the use of `get(value, default)` for an optional attribute.

Beyond the scope of this fix, as we might be introducing more non-breaking functionalities like `reverseIntensity`, we might have to think on how best to codify this knowledge using some formal schema/specification so that others can consume the OMERO.figure format/API and know what to expect.

pwalczysko:
Tested both via editing the json as well as finiding some old figures on eel and verifying that both the edited json as well as the old figure (pre-5.3.0) exports fine. The crash is repeatable on nightshade as well with old figures.

All works as expected. Ready to merge.

omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
jburel: we will also need old testing data in our setup.
I have created https://trello.com/c/h9WOaIkv/10-produce-older-figures-in-the-testing-db
unless we run nightshade clone somewhere later on on

PR #230, opened 2017-06-29
Labels from channel with lut fix

will-moore:
Fixes a bug that has been around since LUTs were supported in OMERO.figure:
See https://trello.com/c/jOCVW3Co/8-bug-create-labels-from-channels-with-luts

To test:
 - Set LUT on a channel
 - Create labels from channels
 - Labels from LUT channels should be grey
 - Export should work as normal


bramalingam:
Works as expected. Looks good to merge.

PR #231, opened 2017-06-30
Fix page_color on multi-page PDF figures

will-moore:
Fixes https://trello.com/c/YiuQbMBL/9-pagecolor-on-multi-page-pdf-export

To test:
 - Create a multi-page figure and set the background colour
 - Export as PDF
 - Check that all pages have the page colour applied


bramalingam:
Works as expected, looks good to merge.

PR #235, opened 2017-07-04
Use CSS rule for padding of image display options instead of hardcoded 2px

carandraug:
While working on adding one more button, I noticed this very small change that would be useful as more buttons get added.


jburel:
--include

jburel:
Thanks
Works as expected merging

PR #236, opened 2017-07-04
Remove unused function arguments and conversion between radians and degrees

carandraug:
While checking how the rotation is done to add a button for horizontal flip, I noticed this unused arguments and extra computations.

### DIFF
--- a/src/js/views/right_panel_view.js
+++ b/src/js/views/right_panel_view.js
@@ -1121,13 +1121,12 @@
                 return {'dx': dx, 'dy': dy};
             }
             var length = Math.sqrt(dx * dx + dy * dy),
-                ang1 = Math.atan(dy/dx),
-                deg1 = ang1/(Math.PI/180);  // rad -> deg
+                ang1 = Math.atan(dy/dx);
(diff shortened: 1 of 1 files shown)


jburel:
Adding the ``include`` flag so it can be part of the daily merge build
--include

joshmoore:
NB: This may well should be considered a breaking change. An alternative would be to replace use of `this.r` with `rotation` within the function method. However, as is always the case with an unused argument, someone with the background should likely evaluate what was really intended.

jburel:
* the conversion changes makes sense
* better to introduce a new function taking 2 arguments and marking the existing ``correct_rotation`` as deprecated.

will-moore:
Since the third argument ```rotation``` wasn't being used, it's not a breaking change to remove that argument. Users could still call it with 3 arguments without an error.
Also, I wouldn't usually consider methods on a View to be an API that could/would be used externally.

jburel:
True that js offers "such flexibility" (not sure it is a good thing!)

jburel:
Thanks

PR #238, opened 2017-08-16
Reverse intensity rename Figure

dominikl:
Simply changes the 'Reverse Intensity' label to 'Invert', to be consistent with the other clients.


mtbc:
Ready for review?

jburel:
We did not review or merge in case we needed a release before 5.4.0
since this will create an indirect dependency on 5.4.0

jburel:
the text is correctly updated. No functionality change
This does not really create a 5.4 dependency from an interaction point of view 
it will create a small naming divergence if used in a 5.3 context.
No need to hold off
Merging

PR #243, opened 2017-09-21
Big images

will-moore:
Adds support for Big Images.

We don't implement a full tiled image viewer, but works by rendering the viewport region each time.
So the experience will not be as good as a dedicated big image viewer (e.g. iviewer).
In general, what needs to be done to make the manipulation acceptable?

 - Add a Big Image to figure / Open_with a big image(s)
 - Zoom into the image, pan around etc. NB: while dragging to pan image, extra pixel data will be missing until drag-stop. Then the panel re-renders with the required region.
 - Rotate and pan / zoom more.
 - Add ROIs - For 'big' images, we allow thick lines (up to 100) and default thickness is 30.
 - Drag and Resize (corner handles and side handles to squash) - ROIs and image should move in sync, then the image re-renders when drag-stop.
 - Export as PDF/TIFF and check it looks same as in web.

TODO: 
 - [x] live update while zooming big images
 - [x] Crop dialog render whole image
 - [x] Figure export - what DPI do we go for in PDFs? Max 300 DPI?
 - [x] Thicker ROI lines? 
 - [ ] Bird's eye view (maybe another PR)
 - [x] Load more pixel data, e.g. rotation not covered - NB: Couldn't get resizing of panel working.
 - [x] Re-render right panel - don't remove old image while new one is loading.
 - [x] Need higher limit on zoom (>1000 %)
 - [ ] When small and Big image are selected together, panning viewport sends small image off screen!
 - [x] Check correct DPI shown for Big images


pwalczysko:
Re:
>Thicker ROI lines?
Yes, if this is not detrimental for small images. I was first confused, thinking the ROIs are not being drawn. Then I made the lines much thicker and red, and the ROIs appeared.

pwalczysko:
I have a very squished feeling from the tiny viewport in the Preview pane. The viewport is even smaller than for a small image. This is probably because my big images are not square. But then, most of the big images we have are not square. Can we have a bigger viewport for big images ?

pwalczysko:
Bug: No rendering settings changes are being respected at all. Is this expected ? In such case disable the rendering controls for big images.

pwalczysko:
Crop does not work for me at all. The image goes to its original size, shape and zoom. No cropping happens.

pwalczysko:
Testing images were http://web-dev-merge.openmicroscopy.org/webclient/?show=image-59301 and http://web-dev-merge.openmicroscopy.org/webclient/?show=image-59307.
Figure saved as ``testing-will-2017``
All user-4, eel merge

will-moore:
Fixed:
 - Rendering settings fixed for big images
 - Crop of big images (test it works for big and small images)
 - Thicker lines available for ROIs on big images (default is also thicker)
 - Rotation now shouldn't show edges of the rendered tile
 - When zooming and panning the image, re-rendering happens without first removing the existing image in the right panel - so you don't see a flash of no-image showing. Much less jarring.
 - Now, right viewport should update interactively while zooming and panning. The max zoom limit is determined by the size of the image. E.g. 70k image can zoom to 30000 %. Can adjust this relationship if needed? Small images are still limited to zoom of 1000 %.

The "squished" viewport is not actually smaller for big images. It is just scaled to fit in the right panel square space, so if the viewport (panel) is very wide and thin, then it will appear thinner in the right panel (size limited by width). If the panel on the figure is set to square, the viewer in right panel will appear bigger.

pwalczysko:
@will-moore 

>E.g. 70k image can zoom to 30000 %. 

I am afraid I do not understand. What is 70% image ? How do you derive the 30000 % from the 70% ?

pwalczysko:
Bug:  side drag handles and corner drag freezes, no image comes up in the ``Preview`` after following sequence:

   - open a big image in Figure http://web-dev-merge.openmicroscopy.org/webclient/?show=image-59301 user-4
   - zoom in to 4452%
   - use the corner drag handles to make the image fairly small, see screenshot 1 below
   - position the image onto the left hand of the canvas
   - use the side control, drag the image long to the right, see screenshot 2 below
   -  observe freeze as described above
Screenshot 1 (after using the corner drag to make image smaller)
screen shot 2017-10-09 at 14 51 39

Screenshot 2 (after using the side handle to make the image wide, all frozen)
screen shot 2017-10-09 at 14 51 49

pwalczysko:
RFE: the dragging of the side handles is resulting in incomprehensible behaviour of the image inside.
After letting the handle go, there is a second of wait, then the image reloads itself, "jumping" a bit towards some direction, not keeping the border areas as they were, it seems pretty random, not sure what is going on.
Was about to suggest to switch the border dragging off, but then, how would the user change the aspect ration ?

pwalczysko:
Stopping testing for now, the points listed in https://github.com/ome/omero-figure/pull/243#issuecomment-334759940 were tested, the problems listed in comments between here and https://github.com/ome/omero-figure/pull/243#issuecomment-334759940

will-moore:
@pwalczysko A 70k image is one that is approximately 70000 pixels on it's longest side. The bigger an image is, the higher the max zoom is.

will-moore:
Fixed:
 - "freeze" bug in right panel
 - "jump" in reloading an image after resizing a big image.

pwalczysko:
both [fixes](https://github.com/ome/omero-figure/pull/243#issuecomment-336876749) work fine.

will-moore:
Export is now working. Ready for testing (I will need to upload script manually first).

src/js/models/panel_model.js
```diff
@@ -546,13 +552,125 @@
...
-            return baseUrl + '/render_image/' + imageId + "/" + theZ + "/" + theT
-                    + '/?c=' + renderString + proj + maps +"&m=c";
+            // If BIG image, render scaled region
+            var region = "";
+            if (this.is_big_image()) {
+                baseUrl = BASE_WEBFIGURE_URL + 'render_scaled_region/';
+                var rect = this.getViewportAsRect();
+                // Render a region that is 1.5 x larger
+                var length = Math.max(rect.width, rect.height) * 1.5;
+                console.log('get_img_src', rect, length, (length - rect.height) / 2);
```
waxenegger: maybe a leftover debug statement?

waxenegger:
works well for me, incl. export as tiff and pdf (attached the output).
[Figure_review_2017-10-19_16:29:15.pdf](https://github.com/ome/omero-figure/files/1397362/Figure_review_2017-10-19_16.29.15.pdf)


When I opened the dev-tools I noticed an error in the console in regards to rafael which complained about a path definition in the svg but I didn't see a corresponding problem in the page. I'll just paste it in here to let you know:
console_log

jburel:
There is similar code to render a region in  ``Figure_To_Pdf.py`` and ``view.py``
It will be better to unify that.

will-moore:
I just uploaded the latest export script to eel.
@waxenegger Sorry - I think you were actually testing the old script but it worked because the image was only 8k * 8k and it was possible to render the whole plane.
I just tested with http://web-dev-merge.openmicroscopy.org/figure/file/31601 and it seems to be working OK...

will-moore:
@jburel Unfortunately there's no way to unify the region rendering code given the current architecture since we need the code on the OMERO.server side for the script AND in the OMERO.web figure app.
This code needs to go in to Blitz Gateway so it is available everywhere. That will take time to make sure it is generic enough and not just OMERO.figure specific, but this will be best since it is useful functionality.
I've creates a card for that https://trello.com/c/Ss88xkZw/102-renderscaledregion-for-figure
However, that shouldn't block this PR.

Alternative for future as we've discussed (https://github.com/ome/omero-figure/pull/240#issuecomment-338232494) is to put export functionality in a separate pip-installable module that can be installed on OMERO.web and OMERO.server.

jburel:
@will-moore still 2 todos listed in the description.
What is the status?

will-moore:
@jburel The bird's-eye-view will be for a future PR. The panning of big+small image together is really an edge-case: The behaviour is kinda logical (selected panels are panned in-sync) - it just doesn't make sense for a user to try and pan a big image and small image to the same viewport location.

will-moore:
Demo'd and discussed this PR with @Gebhard-Stopper.
One issue is what resolution to export the tiffs embedded in the PDF. E.g. if you fully zoom out a big image and export at full resolution, this will be a very large image. So we probably want to down-sample by default, to some max dpi - e.g 300dpi or maybe higher? The user could still choose a higher dpi to export at if desired.
For TIFF export, we will always export at 300 dpi but need to be sure that we're down-sampling (not up-sampling/interpolating).

will-moore:
2 options for dpi settings in PDF export:

 - Have a single target dpi, e.g. 300 dpi and everything is set to this, even if export without resampling would result in a slightly higher resolution e.g. 500 dpi. This would keep a single value to set in UI and mean minimal UI changes. In the dpi dialog we turn this text: "Set a minimum resolution for selected panels. When exported as PDF, panel images below this dpi will be resampled as necessary." to this: "Set an export resolution for selected panels. When exported as PDF, panel images will be resampled to this resolution". 

 - Have a range of acceptable export resolution. E.g. 300 - 600 dpi. This would mean that we didn't down-sample an image that would naturally be at e.g. 500 dpi. But more complex UI.

I'll probably go for option 2 if I can make the UI look OK.
Thoughts?

Gebhard-Stopper:
Option two sounds good. If I understand it correctly it would allow
per-image DPI values (in the pre-defined range).
I guess the pre-defined values will be organised in a global list /
dictionary and then applied per image. Maybe, it would be good if one could
add user-defined DPI values if,
for whatever reason, a super-high-res image is needed.

On the other hand, it might also be good, if option one could be used as an
override to apply the same DPI to all images (as an export-option).


One more thing, this could probably be a feature of a future version.
It would be cool, to have some clue how big a figure actually is (e.g. if I
create my figure on an A4 sheet working specifying coordinates or sizes in
cm would be nice).
Or, as a simple variant you could render a ruler (e.g. like in MS Word)
next to the border of the figure, so one has a visual hint about the
dimension.

will-moore:
Each panel has it's own dpi setting so both options would allow per-panel control of dpi. To apply the same dpi settings across all panels, simply select all panels and set the dpi for all.
The advantage of having a range (min - max) of allowed dpi values means that if a panel would naturally fall inside that range then we don't need to do any resampling/interpolation, and that if it's outside that range then we only need to resample enough to get within the range.
I'll see how that option works out...

Noted the 'ruler' feature request at https://trello.com/c/rX8JMdZ0/150-grid-guidelines-or-ruler-overlay

will-moore:
We now allow users to pick a max export dpi and an optional min export dpi.
The max dpi is only used for BIG images. We pick a suitable tile level so that the region is below this limit.
screen shot 2018-03-21 at 13 42 51

To test, export images as PDF, then can use Illustrator to check panel dpi according to instructions at
http://swainhart.org/how-to-check-the-dpi-of-an-image-in-adobe-illustrator-cs5/

will-moore:
Noticed that crop dialog is rendering the whole image PLUS grey margin, so that we get scroll bars on image and ROI regions are showing wrong region of the image.
screen shot 2018-03-26 at 16 01 32

omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
```diff
@@ -1038,6 +1038,133 @@ def draw_scalebar(self, panel, region_width, page):
...
+                paste_x = -x
+                width = width + x
+                x = 0
+            if y < 0:
+                paste_y = -y
+                height = height + y
+                y = 0
+
+        # Render the region...
+        jpeg_data = image.renderJpegRegion(z, t, x, y, width, height,
```
jburel: handle ``jpeg_data = None``

omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
```diff
@@ -1046,23 +1173,31 @@ def get_panel_image(self, image, panel, orig_name=None):
...
+        size_y = image.getSizeY()
 
         if 'z_projection' in panel and panel['z_projection']:
             if 'z_start' in panel and 'z_end' in panel:
                 image.setProjection('intmax')
                 image.setProjectionRange(panel['z_start'], panel['z_end'])
 
-        pil_img = image.renderImage(z, t, compression=1.0)
+        # If big image, we don't want to render the whole plane
+        if (size_x * size_y) > (4000 * 4000):
```
jburel: should probably retrieve value from configuration

src/js/models/figure_model.js
```diff
@@ -1,7 +1,7 @@
     
     // Version of the json file we're saving.
     // This only needs to increment when we make breaking changes (not linked to release versions.)
-    var VERSION = 2;
+    var VERSION = 3;
```
jburel: new version of the model so that should be a breaking change

src/js/models/panel_model.js
```diff
@@ -528,7 +531,11 @@
                 'y':this.get('y') + (this.get('height')/2)};
         },
 
-        get_img_src: function() {
+        is_big_image: function() {
+            return this.get('orig_width') * this.get('orig_height') > 4000 * 4000;
```
jburel: this should come from config

jburel:
now that the testing infrastructure is in place.
Could you add tests for method changes/new methods?

omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
will-moore: Done in https://github.com/ome/omero-figure/pull/243/commits/d4325e19f6bacc26058c0d411a39fc960d1213a2

will-moore:
--exclude
To allow #253 to be merged

will-moore:
Updated testing scenarios at https://github.com/openmicroscopy/ome-internal/pull/384

Although it wouldn't hurt to test everything (it's been a while since we've done a thorough test) the areas to focus on for this PR are:
 - Panning, zoom, rotation of panels
 - Crop dialog
 - ROI dialog
 - Setting dpi (Info tab)
 - Export as PDF / TIFF

will-moore:
Wanted to start adding more tests, but having problems running tests locally (without ```omero-test-infra```).

```
$ ./setup.py test -k test_list_figures
...
../../OMERO/openmicroscopy/dist/lib/python/omeroweb/webclient/views.py:72: in <module>
    from controller.container import BaseContainer
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

    
>   from webclient.controller import BaseController
E   ImportError: No module named webclient.controller

../../OMERO/openmicroscopy/dist/lib/python/omeroweb/webclient/controller/container.py:31: ImportError
```

However, the export script tests are running OK locally, so I'll add some more of them...

will-moore:
This is now failing to generate pyramids because of the issue fixed in https://github.com/openmicroscopy/openmicroscopy/pull/5720
which is not in the docker image used in ```omero-test-infra```.
cc @jburel I could update the test to use ```self.client``` instead of creating a new user & client, which is what works for pyramid generation in your ```test_rendering.py```.
Need to fix travis to allow this PR to be merged in the build for testing...

jburel:
see https://github.com/openmicroscopy/omero-test-infra/blob/master/.env
The released version is used.

omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
```diff
@@ -1109,6 +1109,140 @@ def draw_scalebar(self, panel, region_width, page):
...
+
+        pil_img = self.render_big_image_region(image, z, t, viewport_region,
+                                               max_width)
+
+        # Optional rotation
+        if rotation != 0 and pil_img is not None:
+            w, h = pil_img.size
+            # How much smaller is the scaled image compared to viewport?
+            # This will be the same 'scale' used in render_big_image_region()
+            scale = float(w) / viewport_region['width']
```
jburel: Could ``viewport_region['width']`` be ``0``?

omero_figure/views.py
```diff
@@ -116,6 +123,83 @@ def img_data_json(request, image_id, conn=None, **kwargs):
     return HttpResponse(json.dumps(rv), content_type='json')
 
 
+@login_required()
+def render_scaled_region(request, z, t, iid, conn=None, **kwargs):
```
jburel: order ``iid, z, t`` will be more logical

omero_figure/views.py
```diff
@@ -116,6 +123,83 @@ def img_data_json(request, image_id, conn=None, **kwargs):
...
+            paste_x = -x
+            width = width + x
+            x = 0
+        if y < 0:
+            paste_y = -y
+            height = height + y
+            y = 0
+
+    # Render the region...
+    jpeg_data = image.renderJpegRegion(z, t, x, y, width, height, level=level,
```
jburel: why not using ``render_image_region``?

src/js/models/panel_model.js
```diff
@@ -478,7 +479,9 @@
                 scale = Math.max(xPercent, yPercent);
 
             // if not zoomed or panned and panel shape is approx same as image...
-            if (dx === 0 && dy === 0 && zoom == 100 && Math.abs(xPercent - yPercent) < 0.01) {
+            var orig_wh = orig_width / orig_height,
+                view_wh = width / height;
```
jburel: could height be 0?

src/js/views/info_panel_view.js
```diff
@@ -0,0 +1,280 @@
+
+var InfoPanelView = Backbone.View.extend({
+
+    template: JST["src/templates/info_panel_template.html"],
+    xywh_template: JST["src/templates/xywh_panel_template.html"],
+
+    initialize: function(opts) {
+        // if (opts.models) {
```
jburel: remove

src/js/views/info_panel_view.js
```diff
@@ -0,0 +1,280 @@
...
+        var aspectRatioStatus = false;
+        this.models.forEach(function(m) {
+            if (attr === 'x' || attr ==='y') {
+                var old = m.get(attr);
+                var coords = {};
+                coords[attr] = old;
+                var offset = this.figureModel.getPageOffset(coords);
+                var newValue = old - offset[attr] + value;
+                // Keep panel within figure limits
+                if (attr === 'x'){
```
jburel: space

src/js/views/info_panel_view.js
```diff
@@ -0,0 +1,280 @@
...
+                    if (newValue > figsize.h || newValue < 0) {
+                        this.ignoreChange = false;
+                    }
+                    newValue = Math.min(figsize.h, newValue);
+                }
+                newValue = Math.max(0, newValue);
+                m.set(attr, newValue);
+            }
+            else {
+                if (value<1) {
```
jburel: spaces

src/js/views/info_panel_view.js
```diff
@@ -0,0 +1,280 @@
...
+                newWidthHeight[attr] = value;
+
+                if ($(".setAspectRatio", this.$el).hasClass("aspectRatioSelected")) {
+                    aspectRatioStatus = true;
+                    var widthCur = m.get('width');
+                    var heightCur = m.get('height');
+                    var aspRatio = widthCur/heightCur;
+
+                    if (attr === 'width'){
+                        var heightNew = value/aspRatio;
```
jburel: could be 0?

src/js/views/info_panel_view.js
```diff
@@ -0,0 +1,280 @@
...
+                var label = v.label || v.id;
+                imageLinks.push({'text': label, 'url': url});
+            });
+        }
+        return imageLinks;
+    },
+
+    // render BOTH templates
+    render: function() {
+        // If event comes from handle_xywh() then we dont need to render()
```
jburel: typo don't

src/js/views/info_panel_view.js
```diff
@@ -0,0 +1,280 @@
...
+
+        if (json) {
+            var html = this.template(json),
+                xywh_html = this.xywh_template(json);
+            this.$el.html(html + xywh_html);
+        }
+        this.rendering = false;
+        return this;
+    }
+});
```
jburel: add extra line

src/js/views/modal_views.js
```diff
@@ -7,28 +7,55 @@
 
         model: FigureModel,
 
+        initialize: function(options) {
+
+            var self = this;
+            // when dialog is shown, clear and render
+            $("#dpiModal").bind("show.bs.modal", function(){
```
jburel: space

src/js/views/right_panel_view.js
```diff
@@ -1072,6 +794,8 @@
                     self.models.forEach(function(m){
                         m.save(to_save);
                     });
+                    // we don't listenTo zoom change...
```
jburel: typo

src/js/views/right_panel_view.js
```diff
@@ -1162,8 +889,12 @@
         // TODO: Update each panel separately.
         update_img_css: function(zoom, dx, dy, save) {
 
-            dx = dx / (zoom/100);
-            dy = dy / (zoom/100);
+            var scaled_dx = dx / (zoom/100);
```
jburel: could it be ``0``?

src/templates/viewport_template.html
```diff
@@ -5,14 +5,5 @@
...
-                style="opacity:<%= opacity %>; left:<%= css.left %>px; top:<%= css.top %>px;
-                    width:<%= css.width %>px; height:<%= css.height %>px;
-                    -webkit-transform-origin:<%= css['transform-origin'] %>;
-                    transform-origin:<%= css['transform-origin'] %>;
-                    -webkit-transform:<%= css['transform'] %>;
-                    transform:<%= css['transform'] %>"
-                src='<%= css.src %>' />
-            <% }); %>
+            <%= inner_template({imgs_css:imgs_css, opacity:opacity}) %>
         </div>
```
jburel: Add extra line

src/js/models/figure_model.js
```diff
@@ -132,6 +132,16 @@
                     }
                 });
             }
+            if (v < 3) {
+                console.log("Transforming to VERSION 3");
```
jburel: to be removed

will-moore: @jburel I've added these console.logs for all the upgrade steps since I think they might help debugging at some point. Don't think this limited logging does any harm.

src/js/views/right_panel_view.js
will-moore: No, the zoom is restricted to be 100% or more.

src/js/views/right_panel_view.js
will-moore: ```listenTo()``` is a Backbone method, so it's referring to the fact that we don't call ```self.listenTo()```.

src/js/views/modal_views.js
will-moore: Where? After the ```function()``` ?
When an anonymous function is passed as an argument like this we don't have a space in ```function(){```. I guess this may not be an official style, and I'm not entirely consistent, it is mostly the style I've adopted. I guess we could try and define/adopt a style guide and clean up a whole bunch of these at once.

omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
will-moore: In theory it's possible to have a panel of zero width or height. In practice it's quite hard to create one (by dragging to resize) but I guess I should make it impossible since it does cause errors long before this point (export).

omero_figure/views.py
will-moore: Do you mean ```image.renderImageRegion()```? That doesn't exist.
```image.renderImage()``` returns a PIL image and we mostly want the ```jpeg_data``` to return in the response.

src/js/models/panel_model.js
will-moore: See above - ensuring that users don't create panels with zero width or height makes the most sense, but should probably go in a separate PR. There's plenty of places that would break in that case, so this PR doesn't add any new bugs for that.

src/js/views/info_panel_view.js
will-moore: See comments above.

src/js/views/right_panel_view.js
jburel: that was not very clear so it will be better to be more explicit

src/js/views/modal_views.js
jburel: this is more for consistency

will-moore:
@pwalczysko I added a spinner that shows on the figure panel while the image is rendering. I couldn't actually make the rendering any faster.
Also fixed the max zoom for non-big images to 1000% as it was before this PR.

omero_figure/views.py
jburel: maybe too much logic in this PR that it cannot be done

pwalczysko:
Sorry, the Projection on a biggish image with multi-z does not work. This is not specific to Figure, iviewer is doing the same thing. The controls work okay, but the "projection" is not a projection, instead, the last section is being displayed by the app.

See images in dataset http://web-dev-merge.openmicroscopy.org/webclient/?show=dataset-23405 login as user-1

cc @will-moore @jburel

jburel:
Card added for projection
https://trello.com/c/HsELkfOJ/14-disable-projection-for-big-images

PR #249, opened 2017-10-20
Update doc link in README

hflynn:
Noticed this while reviewing things for the docs training


jburel:
Change makes sense 
Thanks for spotting that one

PR #252, opened 2017-11-08
Crop page to panels

will-moore:
This allows users to crop a figure to fit around panels.
See https://trello.com/c/lx5YoDTH/14-crop-page-to-panels

To test:
 - With several panels on a figure...
 - Choose File -> Paper setup and pick "Crop page around panels" from the Size options.
 - When OK is clicked, the page should be cropped around the panels, with a small margin.
 - This is added to the Undo queue and can be undone/redone in one action.
 - Subsequent editing of page Size should show that we now have a "Custom" size, and can tweak custom size further.

### DIFF
--- a/src/js/models/panel_model.js
+++ b/src/js/models/panel_model.js
@@ -661,28 +661,28 @@
         },
 
         getAverageWH: function() {
-            var sumWH = this.inject(function(memo, m){
+            var sumWH = this.reduce(function(memo, m){
                 return memo + (m.get('width')/ m.get('height'));
             }, 0);
             return sumWH / this.length;
         },
 
         getSum: function(attr) {
-            return this.inject(function(memo, m){
+            return this.reduce(function(memo, m){
                 return memo + (m.get(attr) || 0);
             }, 0);
         },
 
         getMax: function(attr) {
-            return this.inject(function(memo, m){ return Math.max(memo, m.get(attr)); }, 0);
+            return this.reduce(function(memo, m){ return Math.max(memo, m.get(attr)); }, 0);
         },
 
         getMin: function(attr) {
-            return this.inject(function(memo, m){ return Math.min(memo, m.get(attr)); }, Infinity);
+            return this.reduce(function(memo, m){ return Math.min(memo, m.get(attr)); }, Infinity);
         },
 
         allTrue: function(attr) {
-            return this.inject(function(memo, m){
+            return this.reduce(function(memo, m){
                 return (memo && m.get(attr));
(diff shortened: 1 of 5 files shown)


jburel:
We need to review the board and finish other PRs before moving so we can have a 3.2.0 out

pwalczysko:
Other than https://github.com/ome/omero-figure/pull/252#issuecomment-352457835 the functionality works as described.

will-moore:
@pwalczysko PDF page should be improved now.
Also found and fixed another minor bug. To test:
 - After cropping as above, reopen the Paper setup dialog - note the Custom sizes set (in mm).
 - Save the figure, refresh the browser and re-open Paper setup - check that Custom sizes are same.

src/js/models/panel_model.js
```diff
@@ -628,6 +631,56 @@
             return dpi.toFixed(0);
         },
 
+        getBoundingBox: function() {
```
jburel: This method can be removed

jburel:
* For small figures the script has improved
[from_script (1).pdf](https://github.com/ome/omero-figure/files/1622422/from_script.1.pdf)
but  ideally the displayed url should be the figure_id and when the user clicks on it it opens the figure.
Same for url for the image location

* The feature should be improved for figure on multiple pages and set it automatically to 1 if it is done on figure with more than one page e.g. user-1 http://web-dev-merge.openmicroscopy.org/figure/file/20503

will-moore:
To test last commit:
 - If figure has multiple pages, then 'Crop' option text is now "Crop single page to panels" instead of "Crop page around panels".
 - If the crop option is chosen, the "Number of Pages" field is updated to 1 and clicking "OK" will crop AND set to single page.

I will look at updating PDF handling for small page sizes, but this might be a lot of work and could go in a different PR since it's not directly related to the cropping functionality of this PR.

jburel:
* The text of is correctly modified depending on the number of pages
* if multiple pages, the number of page is correctly set to 1
 * Clicking ``OK``, set the display to 1 page only

@will-moore could you create a card for PDF and small image
and remove the unused method in ``panel_model.js``

jburel:
screenshot in help will have to be updated to include the new option.

will-moore:
Card for small PDF pages created: https://trello.com/c/WZugztMX/172-small-pdf-info-page-formatting
```getBoundingBox()``` removed above.

will-moore:
Help PR: https://github.com/openmicroscopy/ome-help/pull/284

PR #254, opened 2017-11-21
Add labels from tags

will-moore:
Quick PR on the train to SMG 2017....

See https://trello.com/c/ZdmnimZj/64-labels-from-tags

This simply adds a label for every Tag on selected images.

To test:
 - When adding labels, choose ```[tags]``` option.
 - Tags added for each label (only 1 label even if multiple users have added the same Tag).

### DIFF
--- a/src/js/models/panel_model.js
+++ b/src/js/models/panel_model.js
@@ -708,5 +708,38 @@
             }
         },
 
+        createLabelsFromTags(options) {
+            // Loads Tags for selected images and creates labels
+            var image_ids = this.map(function(s){return s.get('imageId')})
+            image_ids = "image=" + image_ids.join("&image=");
+            // TODO: Use /api/ when annotations is supported
+            var url = WEBINDEX_URL + "api/annotations/?type=tag&limit=1000&" + image_ids;
+            $.getJSON(url, function(data){
+                // Map {iid: {id: 'tag'}, {id: 'names'}}
+                var imageTags = data.annotations.reduce(function(prev, t){
+                    var iid = t.link.parent.id;
+                    if (!prev[iid]) {
+                        prev[iid] = {};
+                    }
+                    prev[iid][t.id] = t.textValue;
+                    return prev;
+                }, {});
+                // Apply tags to panels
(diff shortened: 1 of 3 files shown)


src/js/models/panel_model.js
```diff
@@ -708,5 +708,38 @@
             }
         },
 
+        createLabelsFromTags(options) {
+            // Loads Tags for selected images and creates labels
+            var image_ids = this.map(function(s){return s.get('imageId')})
+            image_ids = "image=" + image_ids.join("&image=");
+            // TODO: Use /api/ when annotations is supported
+            var url = WEBINDEX_URL + "api/annotations/?type=tag&limit=1000&" + image_ids;
```
jburel: maybe offering the option later on to change the limit.
this will also depend on the number of images

will-moore: This limit is just a very liberal cut-off to make sure we don't accidentally load millions of annotations, but that scenario is very edge-case.

pwalczysko:
The functionality works as expected. 
Discussing with @will-moore , we went through a round of thinking "what if there are many tags on an image..." but then concluded that in such case, the user would be better off to type the label name manually. The user can also first add all the labels from the tags on the image in question and then delete the labels which do not fit. 

Happy to merge FMPOV.

joshmoore:
Can someone paste screenshots here? Are there any doc screenshots which then need updating?

pwalczysko:
screen shot 2017-12-18 at 16 56 32

pwalczysko:
@joshmoore see screenshot above. I can see two screenshots needing replacing in http://help.openmicroscopy.org/figure.html#labels section. The screenshots are under points 13. and 15.

jburel:
Discussed today @will-moore will open a PR to update the screenshot in help

will-moore:
I was wondering whether to simply add the "Tags" option to fix existing screenshot (above) or to add the whole "Labels from Tags" workflow.
I'll start anyway...

will-moore:
Help PR opened at https://github.com/openmicroscopy/ome-help/pull/283

PR #257, opened 2017-12-18
export_bug_dx_None

will-moore:
See https://trello.com/c/0TnhjDrs/1-export-bug-dx-none

This fixes a rarely seen bug (not sure how to reproduce it in normal use) in figure export.
Check that figure export still works as expected, with some images zoomed in and panned off-centre. Check cropped region is same in exported figure as in web.

### DIFF
--- a/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
+++ b/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
@@ -61,6 +61,7 @@ except ImportError:
     reportlab_installed = False
     logger.error("Reportlab not installed.")
 
+DEFAULT_OFFSET = 0
 
 ORIGINAL_DIR = "1_originals"
 RESAMPLED_DIR = "2_pre_resampled"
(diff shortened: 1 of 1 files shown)


omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
```diff
@@ -774,8 +774,8 @@ def get_crop_region(self, panel):
         zoom = float(panel['zoom'])
         frame_w = panel['width']
         frame_h = panel['height']
-        dx = panel['dx']
-        dy = panel['dy']
+        dx = panel.get('dx', 0)
```
jburel: since you have code after checking if ``if dx > 0`` (same for ``dy``) it will be better to introduce a default value ``default_value= 0``. So if later it changes it is only one place to adjust

will-moore: I can't see that we'd ever want a default offset that wasn't zero.
The default for these offsets is really set in the ```panel_model.js``` and it's only rarely (not reproducible) that it's not set here in the script (so we have to catch it just in case).

jburel: Why 0 is set in ``get``, this is "discovered" further down in another check
having a default value makes that clear and explicit

jburel: could introduce a variable with value 0 I think it will make the logic clearer

will-moore:
@jburel Added default variable.

jburel:
If you could also change ``if dx > 0:`` to ``if dx>DEFAULT_OFFSET`` same for ``dy``

will-moore:
In that case, we really just need to know if ```dx``` is positive or negative and this has nothing to do with the default offset.
E.g. if the default offset was 100, we'd still need to know that ```dx``` is positive (not if it was different from 100).

jburel:
understood 
everything seems to work  now merging

PR #258, opened 2018-01-11
fix theZ display in crop dialog

will-moore:
Fixes https://trello.com/c/imMLEy89/100-bug-figure-crop-dialog-thez-display

To test:

 - Add rectangles on a multi-Z image in Insight, so that the Z range goes from below 100 to over 100 OR from below 10 to above 10.

 - Add the image to a Figure and click on the Crop button to launch the dialog

 - Check that the Z-index ranges are the same as created in Insight.

Before and after shots below:

screen shot 2018-01-11 at 16 15 47
screen shot 2018-01-11 at 16 16 12


jburel:
@will-moore could you please indicate an image with a large number of z-sections on eel?

jburel:
Works as expected
I also checked ROI edit and no problem there

PR #259, opened 2018-01-22
Figure infra

jburel:
Modify travis file to use test-infra
travis file will have to be updated when https://github.com/openmicroscopy/omero-test-infra/pull/3 is merged

PR #265, opened 2018-01-26
Scripts parameters

jburel:
Parameters are not optional as indicated
This PR clearly sets the flag

Discovered while working on https://github.com/ome/omero-figure/pull/259
cc @will-moore 

This could be included in 3.2.0
we can then relax later on but at the moment it is broken


will-moore:
Looks good.

kennethgillen:
Seems this change is related to a stack trace on `ome-demoserver` 

cf https://github.com/openmicroscopy/management_tools/issues/663

PR #274, opened 2018-03-26
svg_resize_on_init

will-moore:
Fixes a small bug noticed on very large, high resolution monitors: https://trello.com/c/BnJ6wTz1/168-big-screens-svg-doesnt-resize-to-fit

To test:
 - Open figure on a large, high resolution monitor
 - Try drag-selection on the canvas (shows dotted selection outline) right down to the bottom-right corner. The drag outline should be able to extend fully down to the corner.


will-moore:
cc @mporter-gre If you could test this on your large monitor, that would be great.

mporter-gre:
I couldn't recreate the bug on my monitor on NS, so I asked Iain to do it on his machine. NS had the bug, but it was fixed on Eel.

pwalczysko:
Michael confirmed that the bug was solved on the big screen. I do not have such screen, but the behaviour on small screen is still okay.

PR #276, opened 2018-04-05
T-index label

will-moore:
See https://trello.com/c/cVYdrcXM/131-no-time-stamp-use-t-index-instead
Adds support for labels showing T-index.
Useful if the image doesn't have timestamp metadata.

To test:
 - Add Timelapse images to figure, with and without Time-stamp metadata.
 - Choose to create label from ```Time (index)``` option (NB: this option will not be disabled in image with no timestamp metadata)
 - Label should show the current T-index for the panel
 - Export of the figure should also show correct T-index label for both images.

### DIFF
--- a/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
+++ b/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
@@ -872,9 +872,9 @@ class FigureExport(object):
 
     def get_time_label_text(self, delta_t, format):
         """ Gets the text for 'live' time-stamp labels """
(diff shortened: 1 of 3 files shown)


jburel:
The manipulation of label needs to be reviewed
if for example as add a label from ``Time (index)``, in the edit row, I have to manually change the type.
If I add a second, I cannot edit the first one

jburel:
I understand the change made in this PR
but I think we should review the current implementation

jburel:
Improvement recorded on the testing sheet.
This can be done in a follow-up PR

PR #277, opened 2018-04-16
save cancel fix

will-moore:
From bug reported at https://docs.google.com/spreadsheets/d/1qdOAv0wUgAVHEtlVroOwkeNUoNwst-APSPFJk6-3Sz8/edit#gid=0

* Click New
* Add image ID 
* File -> Save. Use short-cut. 
* Do not Save. Click cancel
* Save button in the header is not longer active. 
* Cannot save from header any more
* Note that the button does not become active after actions e.g. rotate image

To test, repeat above but the Save button does not become disabled if we Cancel the save.
If save goes ahead, the Save button is disabled in the normal way.

The immediate disabling of the Save button when it's clicked was added to prevent Double-click performing a 2 Save POSTs when saving a pre-saved figure (no "Figure Name" prompt). Now we handle that elsewhere (not in the Save-As path).

### DIFF
--- a/src/js/views/figure_view.js
+++ b/src/js/views/figure_view.js
@@ -411,7 +411,6 @@
                 event.preventDefault();
             }
             this.$saveBtn.tooltip('hide');
-            this.$saveBtn.attr('disabled', 'disabled');
             this.save_figure();
         },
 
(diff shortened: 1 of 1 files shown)


jburel:
Still the same problem. 
Button is disabled after clicking ``Cancel``

jburel:
Where is it handle now?

jburel:
Feature works as expected now
Few points noticed while reviewing this PR.
I have added them to the testing sheet

will-moore:
@jburel I just moved it into the appropriate code block: https://github.com/ome/omero-figure/pull/277/files#diff-013a5e2440a847a8353fff4b62193a28R424

jburel:
sorry I misunderstood the comment

PR #281, opened 2018-04-18
Dont scale line widths on zoom

will-moore:
The problem we find with big images is that the current strategy of scaling stroke-widths of ROI shapes to the image pixels means that when you zoom out, the lines get thinner and are invisible for Big Images, unless you set a stroke width of 100!

Also, the scaling of stroke-widths with images is inconsistent with OMERO.iviewer, so thin lines drawn on a Big Image in iviewer are invisible in figure.

This PR includes https://github.com/ome/shape-editor/pull/11 and means that when we zoom an image, the stroke widths stay the same. NB: since shape-editor is not released yet, I have copied the code over to test. When happy, can revert the TEMP commit and update version.
This means that all shapes on figure panels are shown on the figure at the same stroke-width, even when a panel is resized.
Export script is updated so that exported PDF / TIFF match what is in figure web UI.

When upgrading from older versions, we try to scale the line thickness to appear the same.

We now support line widths of 0.25 and 0.5 (previous minimum was 1).

### DIFF
--- a/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
+++ b/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
@@ -284,7 +284,7 @@ class ShapeToPdfExport(object):
         self.canvas.setStrokeColorRGB(r, g, b)
         self.canvas.setFillColorRGB(r, g, b)
 
-        head_size = (stroke_width * 2) + 20
+        head_size = (stroke_width * 4) + 5
         dx = x2 - x1
         dy = y2 - y1
 
--- a/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
+++ b/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
@@ -475,7 +475,7 @@ class ShapeToPilExport(object):
         y1 = start['y']
         x2 = end['x']
         y2 = end['y']
-        head_size = ((shape['strokeWidth'] * 2) + 20)
+        head_size = ((shape['strokeWidth'] * 4) + 5)
         head_size = scale_to_export_dpi(head_size)
         stroke_width = scale_to_export_dpi(shape.get('strokeWidth', 2))
         rgb = ShapeToPdfExport.get_rgb(shape['strokeColor'])
--- a/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
+++ b/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
@@ -86,6 +86,14 @@ processing steps:
 """
 
 
+def scale_to_export_dpi(pixels):
+    """
+    Origianl figure coordinates assume 72 dpi figure, but we want to
+    export at 300 dpi, so everything needs scaling accordingly
+    """
+    return pixels * 300/72
+
+
 def compress(target, base):
     """
     Creates a ZIP recursively from a given base directory.
@@ -215,8 +223,7 @@ class ShapeToPdfExport(object):
         g = float(rgb[1])/255
         b = float(rgb[2])/255
         self.canvas.setStrokeColorRGB(r, g, b)
-        stroke_width = shape['strokeWidth'] if 'strokeWidth' in shape else 2
-        stroke_width = stroke_width * self.scale
+        stroke_width = shape.get('strokeWidth', 2)
         self.canvas.setLineWidth(stroke_width)
 
         rotation = self.panel['rotation'] * -1
@@ -250,7 +257,7 @@ class ShapeToPdfExport(object):
         g = float(rgb[1])/255
         b = float(rgb[2])/255
         self.canvas.setStrokeColorRGB(r, g, b)
-        stroke_width = shape['strokeWidth'] * self.scale
+        stroke_width = shape['strokeWidth']
         self.canvas.setLineWidth(stroke_width)
 
         p = self.canvas.beginPath()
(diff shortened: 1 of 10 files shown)


omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
```diff
@@ -86,6 +86,14 @@
 """
 
 
+def scale_to_export_dpi(pixels):
+    """
+    Origianl figure coordinates assume 72 dpi figure, but we want to
```
jburel: typo

jburel:
commit after 19e393511843a112d2c955b6dcba92bec89bbca5  will need to be moved to the other PR 
It is possible to split the original PR into few PRs
this is a huge one already and it is probably not finished

will-moore:
Sorry, not sure I understand - can discuss tomorrow...
I'm not sure if the Big Images PR can be split up, since testing any part of it requires other parts.

jburel:
This PR also includes commit from https://github.com/ome/omero-figure/pull/275

will-moore:
Cleaned up history to remove commit from #275.
Last commit handles updating from previous JSON version 2 -> 3 and we scale the strokeWidths of shapes to appear similar, but also ensure that minimum width is 1.
Screenshots show a figure created with v3.2.1 and then opened with last commit.
StrokeWidths on the original are the same for corresponding shapes in the 3 panels, but they appear different widths because of the pane sizes.
In the updated figure, strokeWidths are now different between the 3 panels, because they've been scaled differently in the upgrade. The appearance is similar, except for thin lines that are now no thinner than 1px on the page.

Before (v3.2.1)
screen shot 2018-04-19 at 11 32 45

After (current)
screen shot 2018-04-19 at 11 33 46

will-moore:
Added support for line widths of 0.25 and 0.5 so we can more closely match existing thin lines when upgrading.

pwalczysko:
As discussed with @will-moore , all works fine here, except the gap in sizes causes some arrows which were fine in old versions to appear too thin. This is (hopefully) caused by the gap of the allowed sizes between ``1`` and ``0.5``. @will-moore will add the ``0.75`` size as well, which matches what Illustrator is doing and this will hopefully fix it.

Another ToDo mentioned by @will-moore here is adding the upgrade logic to python, so that the export script can be run on old figures without opening these figures in the Figure UI first. As the opening in the Figure UI automatically upgrades the Figure, we need to take care for the situations where the old figures are directly fed into an export script.

will-moore:
@pwalczysko I added support for 0.75 thick lines.
Also, you can now test exporting of version 2 JSON via running the Figure_To_PDF.py figure script from the webclient scripts menu, pasting in the JSON.

pwalczysko:
Unrelated issue with a dialog, created https://trello.com/c/UVmZ2YsJ/18-bug-typo-in-dialog

pwalczysko:
The 0.75 thickness is there and works nicely and I think is solves the problem

pwalczysko:
Bug?: I do not seem to be able to copy ROIs from a small image to a big one. This means, the figure UI is pretending that I succeeded, and all is fine. But I cannot find the ROIs I just pasted onto the big image on it. I do not think this is a question of color, the ROIs are suitably colored.  Also, I understand that with this PR, the ROIs pasted onto the big image should appear as big as they are on a small one. http://web-dev-merge.openmicroscopy.org/figure/file/41051  username:``outreach``

pwalczysko:
When following the workflow https://github.com/ome/omero-figure/pull/281#issuecomment-388324304, it is important to note that the figure pdf was created successfully using the webclient ``Figure_to_pdf`` script and looks identical to the pdf coming from export via Figure UI+Ui export as pdf.

pwalczysko:
Two (potential) problems still pending

   - https://github.com/ome/omero-figure/pull/281#issuecomment-388317615 (pasting ROIs from small to big)
   - https://github.com/ome/omero-figure/pull/281#issuecomment-388324304 (Figure_to_pdf errors, are these important ?)

will-moore:
The errors you're seeing are just warnings printed to ```stderr``` which you don't normally see when running the script from webclient.
They should be investigated in another PR but shouldn't be an issue for this one.

pwalczysko:
@will-moore explained the [case of lost Rois](https://github.com/ome/omero-figure/pull/281#issuecomment-388336722) and we found them on the Figure finally. Did not realize that the ROIs will shrink on the big image because of the coordinates which are the same, and thus an arrow has infinetisimal length (although the thickness is kept at, say 2pt). This seems to be reasonable. This PR does solve the problem of shrinking of the ROIs out of view completely when these ROIs have been drawn on the same image zoomed out.

I created https://trello.com/c/ni0q2h5c/16-bug-figuretopdf for the problem https://github.com/ome/omero-figure/pull/281#issuecomment-388324304

Ready to merge FMPOV.

jburel:
Recent changes have improved the drawing of arrow (problem reported during testing). It
looks now like an arrow and not a line 
(not tested the pdf yet)

will-moore:
Updated to use the shape-editor 4.0.0 from npm.
Ready for final testing of figures created from previous release, as discussed with @jburel

pwalczysko:
Re-test:

   - test the script for creation of pdfs and tiffs not using UI https://gist.githubusercontent.com/will-moore/d448a7abd57425734aba426f0952eff6/raw/d6d00f49553a67ba811363f51d27ea38bf0b0e3c/omero_figure_export.py
    - test the upgrade of a figure from nightshade, with ROIs of different sizes transported to eel merge via json
    - test ROIs drawn in Big and small images and scaling down of the images

All works fine. Ready to merge.

PR #282, opened 2018-04-23
save before file open

will-moore:
As reported in testing sheet: https://docs.google.com/spreadsheets/d/1qdOAv0wUgAVHEtlVroOwkeNUoNwst-APSPFJk6-3Sz8/edit#gid=0
If a file is renamed and not saved, then File > Open (choose to save changes) doesn't show the new name in the list.
Also, if a New file is unsaved, then File > Open (choose to save new file) doesn't show the new file in the list.
Both these cases should be fixed and the new changes should be shown in the file list, since we now *wait* for the save call to return before we list files (previously, both calls were made asynchronously).

cc @pwalczysko

### DIFF
--- a/src/js/views/figure_view.js
+++ b/src/js/views/figure_view.js
@@ -394,8 +394,8 @@
                     ["Cancel", "Don't Save", saveBtnTxt],
                     function(btnTxt){
                         if (btnTxt === saveBtnTxt) {
-                            self.save_figure();
-                            callback();
+                            // List files after saving current file
+                            self.save_figure({success: callback});
                         } else if (btnTxt === "Don't Save") {
                             self.model.set("unsaved", false);
                             callback();


pwalczysko:
On Mac Chrome and Win7 IE11 I sometimes get additional confirmation window (which I do not see on Mac FF). 

     - create a New file and do not save it, rename it
     -  File > Open
     - Select "Save"
     - additional dialog pops up, see screenshots below
     - click okay and observe that when you do File > Open, the new name is there in the list as expected
IE11:
ie11-additionaldialog

Mac Chrome
screen shot 2018-04-25 at 15 13 58

jburel:
The extra dialog has been there from the start. 
We could review that later on depending on feedback.
The changes work as expected. Merging

PR #285, opened 2018-05-15
Z-projection disabled big images

will-moore:
Since BlitzGateway methods for rendering big images don't yet support Z-projection, we disable the Z-projection button if any selected images are Big.

To test:
 - Need a Big Image with Z-stack.
 - Whenever any Big image is selected, Z-projection button should be disabled.

cc @jburel


rgozim:
Working fine!

PR #288, opened 2018-05-16
Fix Arrow strokeWidth for PDF export

will-moore:
Missed this from before.

To test:
 - draw arrows of various strokeWidths
 - export as PDF
 - check that stroke-widths look correct compared with web view

Screenshot shows web (left), PDF before fix (centre) and PDF after fix (right)

screen shot 2018-05-16 at 14 38 55


jburel:
Change looks good
Merging

Issue #292, opened 2018-05-18
failure to handle panels with null values for dx and dy


carandraug:
We have a figure where one of the panels has a dx and dy value of
`null`. This makes `FigureExport.build_figure` method to fail like so:

      File "Figure_To_Pdf.py", line 801, in build_figure
        self.add_panels_to_page(panels_json, image_ids, page)
      File "Figure_To_Pdf.py", line 1586, in add_panels_to_page
        image, pil_img = self.draw_panel(panel, page, i)
      File "Figure_To_Pdf.py", line 1408, in draw_panel
        pil_img = self.get_panel_image(image, panel, orig_name)
      File "Figure_To_Pdf.py", line 1326, in get_panel_image
        cx += dx
    TypeError: unsupported operand type(s) for +=: 'long' and 'NoneType'

Seems it goes like this:

1. panels have a `dx` and `dy` value of `null`
2. python's json module converts those values to `None`
3. those values are accessed internally like `panel.get('dx', DEFAULT_OFFSET)`
4. because those keys do exist in the key, the `DEFAULT_OFFSET` is never used

I'm using the `Figure_To_Pdf` script like so:

    from Figure_To_Pdf import FigureExport
    ann = conn.getObject("FileAnnotation", id)
    json = "".join(ann.getFileInChunks())
    script_params = {
        'Figure_JSON' : json,
        'Webclient_URI': 'https://omero1.bioch.ox.ac.uk/webclient/',
        'Export_Option' : 'PDF',
    }
    fig_export = FigureExport(conn, script_params, export_images=False)
    fig_export.build_figure()

A possible solution would be to have `json.loads` ignore values with a
`null` value (possible via the `object_hook` argument).

For what is worth:

* this seems to only happen when using the script programatically.
  When using the 'Export PDF' button on the omero figure application
  it works.  I'm guessing something else on the GUI is fixing the
  `null`s.  That something else is also doing it when choosing the
  'File > Export as JSON' menu.  So maybe that thing could be in the
  omero_figure package?

* I'm not sure how this image ended up with a `null` value there.  We
  had a list of 350 figures to convert, and only one of them had this
  issue.  So maybe this is not an issue on reading the JSON, but on
  how that JSON got generated in first place.

carandraug:
> As noted in #257 the unset dx/dy happens rarely and it's not clear why. Seems we need a more robust solution when running exporting directly instead of via web.

[...]

PR #294, opened 2018-05-21
fix invalid panel dx/dy offsets

carandraug:
After parsing the JSON string, check if it's valid and fix it where we can.  Fixes issue #292


will-moore:
I've not tested yet but looks good.
We will be releasing figure 4.0.0 very soon so this won't go into that release I'm afraid.
I just noticed travis failed:
```
./omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py:651:9: E266 too many leading '#' for block comment
./omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py:652:9: E266 too many leading '#' for block comment
```

PR #297, opened 2018-06-07
Export figure no script

will-moore:
This separates the figure export functionality from the OMERO.script so that we can export figures without running any scripts.

TODO: Need to decide how to distribute the ```export``` module. The script needs this on the server, and it could also be useful independently of the web app so it should probably be it's own ```pip installable``` app. Does this have to go in it's own separate repo?

Until then, the export script will not work:

--exclude

The export returns a file_object, without writing any images or thumbnails to disk (the export script previously wrote files to current folder since it was a temp script folder).

Example usage to save the PDF / TIFF or ZIP to disk:
```
from omero_figure.export import FigureExport

fig_export = FigureExport(conn, script_params)
file_object = fig_export.build_figure()
file_data = file_object.getvalue()
file_object.close()
file_name = fig_export.get_export_file_name()
with open(file_name,'wb') as out:
        out.write(file_data)
```

For export to new Image in OMERO (or multiple images for multi-page figure), we get a list of new Image IDs:
```
fig_export = OmeroExport(conn, script_params)
image_ids = fig_export.build_figure()
```

This makes many of the same changes as https://github.com/ome/omero-figure/pull/240 but we still use the Scripting Service to do the exports for the web app so the behaviour is unchanged.

cc @jburel @joshmoore @carandraug


will-moore:
As expected, the export script is failing in travis integrations tests, since the export module is not included on the server:
```
E         File "./script", line 26, in <module>
E           from omero_figure.export import FigureExport, \
E       ImportError: No module named omero_figure.export
```

will-moore:
cc @jburel need to discuss the next steps here

PR #300, opened 2018-06-25
fix render_scaled_region for Pillow 4.2

will-moore:
This fixes this issue https://github.com/python-pillow/Pillow/issues/2609 reported by @pwalczysko using ```pillow==4.2.1```.

To test, add a Big Image to figure and check that it renders OK.


pwalczysko:
This change works for me on the localhost with pillow 4.2.1

jburel:
Has this been tested with other version of Pillow?

jburel:
Does the problem only occur when generating a jpg?

will-moore:
There's nowhere else in views.py or Export script that we're saving PIL image as JPEG, so nowhere else needs fixing.
However, I made the same ```RGB``` change at views.py ```def default_thumbnail``` in case this changes to saving as JPG. Functionality is unchanged.

jburel:
Thanks @will-moore

PR #301, opened 2018-07-04
Fix dropdown buttons on Windows

will-moore:
Hopefully fixes bug on Windows:
https://www.openmicroscopy.org/community/viewtopic.php?f=4&t=8534

### DIFF
--- a/src/js/views/right_panel_view.js
+++ b/src/js/views/right_panel_view.js
@@ -269,8 +269,9 @@
             }
             // All others, we take the <span> from the <a> and place it in the <button>
             if ($span.length === 0) $span = $a;  // in case we clicked on <span>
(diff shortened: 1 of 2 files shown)


mtbc:
If it does fix it we can then comment on the forum thread accordingly. :+1: for the explanatory comment.

jburel:
I have tested latest and the merge with Firefox, Chrome and IE on Windows
and I do not see any difference
I might be missing something.

will-moore:
Tested on the Windows machine where I reproduced this earlier and it is fixed with this PR. 

NB: I also realised that this bug is due to touch-screen functionality which is why we only see it on *some* machines. When touch-screen is disabled then we don't see the bug.

PR #302, opened 2018-07-05
Default image viewer is figure, or iviewer

will-moore:
This is an idea that could improve the workflow where we use Images of a figure to organise figures in OMERO (see https://github.com/openmicroscopy/design/issues/96).

If we configure the default viewer of webclient to be OMERO.figure:
```
$ bin/omero config set omero.web.viewer.view omero_figure.views.figure_image_viewer
```
now, when an image is double-clicked in webclient, we test to see if it has been exported from Figure (does it have the OMERO.figure URL in description)? If so, we redirect to open the original figure in OMERO.figure.
Otherwise we simply open the image in iviewer as normal (or webgateway viewer if iviewer not installed).

To test:
 - Setup config as above
 - Save a figure and export the figure as "OMERO Image"
 - Follow link to that image in webclient (link shown when export completes)
 - Double-click on that Image in webclient (or click "Full Viewer" button)
 - It should open the figure in OMERO.figure app.
 - Other images in OMERO should open in iviewer as normal.

cc @jstitlow

### DIFF
--- a/omero_figure/views.py
+++ b/omero_figure/views.py
@@ -16,12 +16,13 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 
-from django.http import Http404, HttpResponse
+from django.http import Http404, HttpResponse, HttpResponseRedirect
 from django.shortcuts import render
 from datetime import datetime
 import unicodedata
 import json
 import time
+import re
 
 from omeroweb.webgateway.marshal import imageMarshal
 from omeroweb.webgateway.views import _get_prepared_image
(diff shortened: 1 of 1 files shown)


jburel:
This is an interesting idea. A problem i can see with the current strategy is that if that if the "default" viewer is not iviewer e.g. pathviewer, the image will pop up in the old gateway viewer.

will-moore:
@jburel Yes, that's always going to be a problem with any conditional opening of an image viewer. E.g. if 3D image open in FPBioimage else open in....?
I guess you could have 2 views.py methods ```figure_image_viewer_or_iviewer``` and ```figure_image_viewer_or_pathviewer``` and then the config above could choose which you want ```$ bin/omero config set omero.web.viewer.view omero_figure.views.figure_image_viewer_or_*```

PR #303, opened 2018-07-06
Z slider bugfixes

will-moore:
See https://trello.com/c/zXvu3QjV/202-bug-z-slider-bugs

To test:
 - Clicking Z / T increment buttons should update Z/T labels and the slider position.
 - Selecting the Z/T slider handle, then using the arrow keys to increment the slider should not nudge the selected panels in the figure.
 - Check other Z/T slider & label functionality is unchanged: Test images with different Z/T sizes and values.


jburel:
Tested the changes
Everything works as expected.
Display of time using unit will need to be reviewed.
e.g. an image with 6 timepoints show 0 secs  all the time.
This is outside the scope of this PR

PR #304, opened 2018-07-09
big image scalebar fix

will-moore:
Bug fix: See https://trello.com/c/S78nkV0c/203-bug-scalebars-in-exported-pdf-big-images (reported at https://www.openmicroscopy.org/community/viewtopic.php?f=4&t=8523&p=19515)

To test:
 - Check that scalebars are drawn correctly at various zoom levels of a big image.
 - I did this by drawing ROI lines of known length in iviewer, using the scalebar on the image.
 - Then add these Lines as ROIs in figure and add scalebars of corresponding length, e.g. screenshot below:
 - Add some scalebars to non-big images to check they still work OK.
 - Export to PDF & TIFF and check that scalebars show up at same level as in web.

screen shot 2018-07-09 at 12 51 10

### DIFF
--- a/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
+++ b/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
@@ -1118,7 +1118,16 @@ class FigureExport(object):
             ly = y + height - spacer
             align = "right"
 
-        pixels_length = sb['length'] / panel['pixel_size_x']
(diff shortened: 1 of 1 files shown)


rgozim:
Seems okay to me.

jburel:
Having the unit only hard-coded in microns is far from ideal for big images.
Adding a card since we should fix that

PR #306, opened 2018-07-10
Additional export functionality

knabar:
This PR includes additional functionality for the PDF/TIFF export:

- Shape fills (fully transparent when not given)
- Alpha channel for stroke and fill colors (defaulting to fully opaque when not given)
- Rotated rectangles (by processing rectangles as a polygon)
- Point shapes (as small circles processed as ellipses)
- Shape labels

None of these changes should impact the current export behavior for a given figure, though I need to do some more thorough testing.

Especially shape label rendering will need some more work, but opening the PR now for early review.

### DIFF
--- a/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
+++ b/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
@@ -26,6 +26,7 @@ import os
 from os import path
 import zipfile
 from math import atan2, atan, sin, cos, sqrt, radians
+from copy import deepcopy
 
 from omero.model import ImageAnnotationLinkI, ImageI
 import omero.scripts as scripts
@@ -141,6 +142,8 @@ class ShapeToPdfExport(object):
                     self.draw_polygon(shape)
                 elif shape['type'] == "Polyline":
                     self.draw_polyline(shape)
+                elif shape['type'] == "Point":
+                    self.draw_point(shape)
 
     @staticmethod
     def get_rgb(color):
@@ -400,6 +403,12 @@ class ShapeToPdfExport(object):
 
         # Restore coordinates, rotation etc.
         self.canvas.restoreState()
+
+    def draw_point(self, shape):
+        s = deepcopy(shape)
+        s['radiusX'] = 5 / self.scale
+        s['radiusY'] = 5 / self.scale
+        self.draw_ellipse(s)
 
 
 class ShapeToPilExport(object):
--- a/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
+++ b/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
@@ -637,7 +637,7 @@ class ShapeToPilExport(ShapeExport):
         temp_draw = ImageDraw.Draw(temp_image)
 
         # if fill color, draw polygon without outline first
-        if rgba[3]:
+        if closed and rgba[3]:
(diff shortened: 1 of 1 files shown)


jburel:
Comparing both web-latest and web-merge
No obvious regression in the exported figure
I noticed a problem exporting as TIFF see screenshot (left export, right figure)
export_as_tiff

omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
```diff
@@ -357,27 +431,37 @@ def draw_polygon(self, shape, closed=True):
...
+
+        self.draw_shape_label(shape, Bounds(*points))
 
     def draw_polyline(self, shape):
         self.draw_polygon(shape, False)
 
     def draw_ellipse(self, shape):
         stroke_width = shape['strokeWidth']
         c = self.panel_to_page_coords(shape['x'], shape['y'])
+
```
jburel: that means that we wont' be able to handle shapes covering partially the image

will-moore: No, that's always been the case with PDF export. Shapes that extend outside of the cropped image panel are drawn on the page without being confined to the panel that they come from. Haven't found a way to be able to work around that.

jburel: Thanks for the clarification @will-moore

will-moore:
Seems that the issue @jburel noticed above is due to this PR. Tried exporting same figure with script from origin/master (left) and then from this branch (right). NB: also noticed a truncation of the purple Polylline.

screen shot 2018-08-06 at 10 28 09

knabar:
The truncation is due to the large stroke width and the temporary canvas used to draw on not having enough buffer around the shape. Should be easy to resolve once I get back.

knabar:
Rebased on top of 4.0.1 and fixed the offset/cropping issue when drawing polygons

jburel:
@knabar is it ready for review?

knabar:
@jburel Yes, apologies for not explicitly stating so

will-moore:
This has fixed the issue we were seeing above. The only problem I'm seeing now is that the rectangle corners are rounded in TIFF export because of the way that we're drawing polygons.

For rotated rectangles (which we don't yet support in the web UI) it may be harder find a workaround so I guess it's OK to use the polygon method, but if there's no rotation then I think it would be nice to preserve the current behaviour.

I also tested with a rotated image (not rotated rectangles) and it's working fine:

screen shot 2018-09-17 at 10 32 47

knabar:
@will-moore Thanks.  I'll check out the rectangle issue to see if there's an easy fix for both unrotated and rotated.

knabar:
Apparently yesterday a PR was merged into Pillow to draw nicer polyline joints, but they chose rounded joints also: https://github.com/python-pillow/Pillow/pull/3250

knabar:
@will-moore @jburel I changed rectangle drawing for Pillow back to a custom method to get squared corners.  Ready for another review.

screen shot 2018-09-17 at 17 45 50

will-moore:
This is working nicely for me now.
But travis is complaining:
```
./omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py:669:5: E303 too many blank lines (2)
```
Otherwise, good to merge.

knabar:
@will-moore Fixed the blank line

will-moore:
This is also working on web-dev-merge now too. Good to merge.

PR #308, opened 2018-08-22
Don't show NaN for shape Z/T if undefined

will-moore:
Don't show NaN for shape Z/T if undefined.

To test:
 - Load ROIs with multiple shapes where Z / T are not defined.
 - Shapes should not show Z / T as NaN (should be blank)

cc @jburel


will-moore:
Test image at user-3 http://web-dev-merge.openmicroscopy.org/webclient/?show=image-76304

PR #329, opened 2019-04-17
export to omero description

will-moore:
This improves the description of OMERO images created from figure export.
The image description now include references to the images that are contained in the figure (same as the PDF info page does for other forms of figure export).
This helps keep track of where an image-of-a-figure comes from.

To test:
 - export a Figure to OMERO Image
 - check the description for listing the images in the figure
 - In Insight, the ```Image ID: 123``` should be rendered as a clickable link.

### DIFF
--- a/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
+++ b/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
@@ -2214,6 +2214,16 @@ class OmeroExport(TiffExport):
         if legend is not None:
             description = "%s\n\n%s" % (description, legend)
 
+        img_ids = set()
+        lines = []
+        for p in self.figure_json['panels']:
+            iid = p['imageId']
+            if iid in img_ids:
+                continue    # ignore images we've already handled
+            img_ids.add(iid)
+            lines.append('- Image ID: %s %s' % (iid, p['name']))
+        description += "Contains images:\n%s" % "\n".join(lines)
+
         np_array = numpy.asarray(self.tiff_figure)
         red = np_array[::, ::, 0]
         green = np_array[::, ::, 1]


jburel:
* Image's ID and name are displayed.
* The UI could quickly become very clutter. 
So maybe we should display the ID only and name in the tooltip.
Best will be to be able to open the image in iviewer from the ID

Note that insight link is no longer active.

will-moore:
Unfortunately we don't have the ability to use tooltips etc in the description and if we try to truncate the image name then info will be lost and you could end up with ambiguous names.
I think it's better to keep the full name and risk a little clutter than to remove names.

Screen Shot 2019-07-16 at 12 17 15

will-moore:
You can also improve the readability in the screenshot above by resizing the right panel so that image names don't wrap onto a second line.

jburel:
Looking at the screenshot  I am wondering if we should now have
Under Contains Images a list of hyperlinks to the images similar to the link to the figure

will-moore:
I tried manually editing the description above to see how this looks with links. I tried adding links *after* the ```Image ID: 111``` for the first few images, and *instead* of it for the last few images:

Screen Shot 2019-07-17 at 23 45 11

Not sure if we want all this?
Best would be to have ```Image ID: 111``` be rendered as a link, so maybe we should go for this feature and not add other links that will then be redundant.

jburel:
Having Image ID rendered as link is the best option.
It has been turned off in insight (I cannot remember why)

will-moore:
OK, so I'll leave it as it is for now with ```- Image ID: 123 name```.

jburel:
How long will it take to render the ID as link?

will-moore:
Done https://github.com/openmicroscopy/openmicroscopy/pull/6087

Screen Shot 2019-07-18 at 22 59 44

jburel:
Not possible to test the export option on merge-ci, the control is greyed out: script is not installed

jburel:
Retested again
It works as expected

PR #332, opened 2019-07-17
Release travis

jburel:
This PR fixes recent issue with build and adds the option to release to pypi from travis

Check that travis remains green


joshmoore:
(Would it be useful to have a dockerfile here as well?)

jburel:
there is already one for demo

jburel:
@sbesson happy with the change. We will probably release a new version in the coming week

PR #333, opened 2019-07-19
Channel labels in preview tab

will-moore:
See https://trello.com/c/QgzoGsmi/195-channel-names-in-preview-panel

Work in progress...

Screen Shot 2019-07-19 at 16 37 06


jburel:
Ideally the color of the text should also be changed depending on the color of the channel.
Black text on dark channel color is not readable.

will-moore:
I've used the same logic we do in the webgateway viewer (originally from Insight) for showing white text if colour is dark.
Knowing where to truncate long labels with ```...``` is tricky, depending on which letters you have and you end up showing less of the label than you otherwise would. Full label is shown with tooltip.

jburel:
Discussed technical option with @will-moore for the text
The color is now changed according to the background of the button.
Similar changes should be applied to iviewer

will-moore:
Testing options for handling long channel names:
Default view:
Screen Shot 2019-08-20 at 10 55 48

With css ```text-overflow: ellipsis;``` you lose too much info
Screen Shot 2019-08-20 at 11 03 01

Shrinking text to 80% makes it less readable and messes up the height of the buttons.
Screen Shot 2019-08-20 at 11 03 39

Conclusion: best to leave it as it is I think.

will-moore:
cc @dsudar Best way to handle long channel names?

dsudar:
Hi @will-moore 
Great that you're addressing this. I quickly checked how iviewer does things. Same as your "default view" and then a tooltip with the full name when you hover. That actually works quite well and would have my preference. And I agree about the text color with @jburel.
Thanks,
Damir

jburel:
Thanks @will-moore for investigating, the text display at 80% is too small.
So probably better to keep it as it is and have the full name in the tooltip

will-moore:
Great. We already have the full channel label in tooltip so I think this is good to go.

src/js/views/channel_slider_view.js
```diff
@@ -61,6 +61,58 @@ var ChannelSliderView = Backbone.View.extend({
...
+        if (color.endsWith('.lut')) {
+            return false;
+        }
+        var c = this.hexToRgb(color);
+        var min, max, delta;
+        var v, s, h;
+        min = Math.min(c.r, c.g, c.b);
+        max = Math.max(c.r, c.g, c.b);
+        v = max;
+        delta = max-min;
```
jburel: if min = max, then delta will be 0, leading to error below

will-moore: It won't actually lead to an error since a ```number/0``` returns ```Infinity``` in JavaScript:
```
>>> h = 2 / 0
Infinity
>>> h = h * 60
Infinity
>>> h = h/360
Infinity
>>> h > 0.6
true
```

jburel: still a division by 0 is not correct. I will not rely on the interpretation of JS, this could change    and maybe lead to error

will-moore: Fixed in fcf8fa4

PR #338, opened 2019-08-23
Add upgrade script info to README

will-moore:
The README didn't previously explain the need to update the script and several users have reported bugs due to not updating the script when they upgrade figure. Latest example is https://forum.image.sc/t/error-when-exporting-pdf-from-omero-figure-class-s-instances-can-only-be-saved-once-self-class-name/28915/4

This adds upgrade instructions to the README.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -95,6 +95,28 @@ Now install the script's dependencies:
 
     $ yum install python-markdown
 
+Upgrading OMERO.figure
+----------------------
+
+After upgrading OMERO.figure with:
+
+::
+
+    $ pip install -U omero-figure
+
+You need to update the Figure export script using one of the 2 options described
+above. If using *Option1*, you need to *replace* the existing script:
+
+::
+
+    # Get the ID of the existing Figure_To_Pdf script:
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -95,6 +95,28 @@ Now install the script's dependencies:
...
+----------------------
+
+After upgrading OMERO.figure with:
+
+::
+
+    $ pip install -U omero-figure
+
+You need to update the Figure export script using one of the 2 options described
+above. If using *Option1*, you need to *replace* the existing script:
```
mtbc:
```suggestion
above. If using *Option 1*, you need to *replace* the existing script:
```

README.rst
```diff
@@ -95,6 +95,28 @@ Now install the script's dependencies:
...
+----------------------
+
+After upgrading OMERO.figure with:
+
+::
+
+    $ pip install -U omero-figure
+
+You need to update the Figure export script using one of the 2 options described
+above. If using *Option 1*, you need to *replace* the existing script:
```
jburel: Maybe indicate that the user must have some admin privileges to do such operation.

PR #339, opened 2019-08-30
Remove id from image link description

will-moore:
This uses "Image:123" instead of "Image ID: 123" in the description of images created from figures.
As suggested (and supported) in https://github.com/ome/openmicroscopy/pull/6087#issuecomment-517166063

NB: That PR is not merged just now, so links aren't rendered, but layout looks cleaner:

Screen Shot 2019-08-30 at 11 17 41

### DIFF
--- a/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
+++ b/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
@@ -2221,7 +2221,7 @@ class OmeroExport(TiffExport):
             if iid in img_ids:
                 continue    # ignore images we've already handled
             img_ids.add(iid)
-            lines.append('- Image ID: %s %s' % (iid, p['name']))
+            lines.append('- Image: %s %s' % (iid, p['name']))
         description += "Contains images:\n%s" % "\n".join(lines)
 
         np_array = numpy.asarray(self.tiff_figure)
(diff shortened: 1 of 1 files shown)


joshmoore:
> NB: That PR is not merged just now, so links aren't rendered, but layout looks cleaner:

Similar to https://github.com/ome/scripts/pull/152#issuecomment-525635552, we need to consider how this kind of statement will be communicated to users. e.g. someone who updates figure without updating the related server version will _lose_ their links.

will-moore:
The rendering of ```Image: 123``` to html link is not an essential feature from OMERO.figure point of view. The adding of Image references to the Description field was more about keeping track of the data provenance. So that will be the focus of the Figure release announcement.

The rendering of links has not previously been supported in webclient, so no one will *lose* their links by not using https://github.com/ome/openmicroscopy/pull/6087.

joshmoore:
> The rendering of Image: 123 to html link is not an essential feature from OMERO.figure point of view...

Fair enough.

> The rendering of links has not previously been supported in webclient, so no one will lose their links

Ok. But perhaps a surprise for insight users. Mostly, I'm trying to stress that for any of these loosely-coupled dependencies (scripts, text-based formats, etc) we need to have solid strategies for dealing with breaking changes.

will-moore:
OK, I guess I'm fine with it either way.
Image:123 is nice because it's slightly more concise (easier to remember and not make typos) and the same format is supported in CLI etc.
Does support for that need back-porting to the "deprecated" Insight?

jburel:
The fact that it unifies the passing of argument is a valid point CLI, figure etc. will then follow the same pattern.
This means that an adjustment should be made in insight and we will have to handle both space/no space in the UI

jburel:
See insight adjustment https://github.com/ome/omero-insight/pull/78

will-moore:
webclient PR to render links is now available for testing https://github.com/ome/omero-web/pull/16

jburel:
tested  against
https://merge-ci.openmicroscopy.org/web/webclient/?show=image-91366
with and without ID in the description

PR #359, opened 2019-12-20
run NPM before build_py

will-moore:
Port https://github.com/ome/omero-parade/pull/71 from parade...

To test:
 - Check py3-ci has figure working


joshmoore:
https://github.com/ome/omero-figure/blob/7d5bb3155eff68381c1ccd775acef6029e029d07/.omeroci/app-deps runs
```
npm install -g grunt
```
Is that needed?

will-moore:
Even with grunt installed globally, it still fails with ```Fatal error: Unable to find local grunt.``` in ```pip install dist/omero-figure-4.2.dev1.tar.gz ```.
However, the previous ```python setup.py sdist``` seems to compile js OK with grunt, ```File omero_figure/static/figure/figure.js created.``` etc.

We shouldn't really need grunt in the ```pip install``` step.

will-moore:
Even with the grunt install step, still not being found. In travis failure above...

```
  npm install
  ...
  npm install -g grunt
  /usr/bin/grunt -> /usr/lib/node_modules/grunt/bin/grunt
  /usr/lib
  `-- grunt@1.0.4
  
  grunt jst
  grunt-cli: The grunt command line interface (v1.2.0)
  
  Fatal error: Unable to find local grunt.
```

setup.py
```diff
@@ -108,87 +105,22 @@ def run_tests(self):
...
-
-cmdclass['develop'] = Develop
+def require_npm(command, strict=False):
+    """
+    Decorator to run NPM prerequisites
+    """
+    class WrappedCommand(command):
+        def run(self):
+            if strict or not os.path.isdir(
+                    'omero_figure/static/figure/js'):
```
manics:
```suggestion
                    'omero_figure/static/figure'):
```
Looks like the files aren't under `js`:
```
        omero_figure/static/figure/figure.js
        omero_figure/static/figure/templates.js
```

manics: Actually that won't work, you'll need to check for the presence of the files not the directory

setup.py
```diff
@@ -108,87 +105,22 @@ def run_tests(self):
...
+def require_npm(command, strict=False):
+    """
+    Decorator to run NPM prerequisites
+    """
+    class WrappedCommand(command):
+        def run(self):
+            if strict or not os.path.exists(
+                    'omero_figure/static/figure/figure.js'):
+                self.spawn(['npm', 'install'])
+                self.spawn(['sudo', 'npm', 'install', '-g', 'grunt'])
```
manics: Can you take this out completely? I don't think we should install anything globally as part of the build process, if it's required it should be documented.

PR #360, opened 2020-01-08
omero-web dependency

jburel:
add ``omero-web`` as dependency


mtbc:
Could also mention OMERO *.web* in README but fine either way.

PR #366, opened 2020-02-13
Omero figure cli

will-moore:
This aims to simplify the install of figure by registering figure as a CLI plugin to handle setting of config and upload or replace of the figure export script.
This should make the install as easy as:

```
pip install omero-figure
omero figure install
```

The ```install``` command does configuration ```omero config set ....``` for all OMERO.figure config and also uploads the export script.

If the settings are already configured, will not duplicate settings, but if a similar (but different) setting is already present, then the new settings will get added to the list.
E.g. if ```top_link``` title (tooltip) is different then the config will not be recognised as already been set, so the new one will be added to top_links. Same for open_with.

The script upload will replace existing script if already installed, otherwise it will install new script.

I had to move ```omero_figure``` under ```src``` along with ```omero.plugins``` (same layout as for ```omero_metadata```).


will-moore:
Hmm - travis is failing the same as it did in https://github.com/ome/omero-figure/pull/359
```
  Fatal error: Unable to find local grunt.
```
I thought it might be the incorrect path that was causing the problem in #359 but https://github.com/ome/omero-figure/pull/366/commits/512445540248060218f838d522fa123331b9f518 didn't fix it.

cc @manics Any ideas?

jburel:
@will-moore I have pushed a tmp commit that should fix the problem
It requires a change in test-infra

jburel:
Excluding this PR so we go over the ones first

will-moore:
Updated https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-web/configure to use ```omero figure install``` instead of manual config.

manics:
This might be for a separate PR: Since you're refactoring this would it be worth moving most of the functions in https://github.com/ome/omero-figure/blob/15b696f259ea0718c4f70533d6cfe5a8ebcbebbf/src/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py to a Python library e.g. `src/omero_figure/library.py` so the actual `Figure_To_Pdf.py` would be a much simpler wrapper?

mtbc:
`./src/omero/plugins/figure.py:24:1: F401 'sys' imported but unused`

will-moore:
Running 
```
omero figure install
```
on OMERO.web job failed with
```
12:14:17 + omero figure install
12:14:18 stdin is not a terminal: cannot request server
```

joshmoore:
It's trying to login to upload the script? If so, you'll need to proceed with `omero login ... ` or pass all the parameters with the one command.

will-moore:
Where can I get login credentials?

joshmoore:
The default omero-test-infra coordinates would be:
```
omero login root@omero -w omero
```

will-moore:
At this point, OMERO-web job deploy is working with
```
pip install git+git://github.com/$SPACE_USER/omero-figure.git@$MERGE_PUSH_BRANCH#egg=omero-figure
omero login root@omero -w omero
omero figure install
```
https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-web/428/
and the Figure_To_Pdf.py export script is installed and works. 👍 

TODO:
 - Use 'data files' to include the script, instead of within the ```omero_figure``` web app. https://packaging.python.org/guides/distributing-packages-using-setuptools/#data-files
 - Load the script with
```
from pkg_resources import resource_string
...
script_text = resource_string(__name__, path_to_script)
```
instead of utils.read_file()

will-moore:
Initial testing of ```script_text = resource_string(__name__, 'Figure_To_Pdf.py')``` to load the script from the same dir as the plugin 
(as in e.g. at https://github.com/ome/scc/blob/master/scc/version.py#L35):
Resulted in
```
"Can't perform this operation for loaders without 'get_data()'"
NotImplementedError: Can't perform this operation for loaders without 'get_data()'
```
@joshmoore hinted "You’ll probably need to search scc’s setup.py for 'data'", but that only finds ```include_package_data=True,``` which we already have set.

will-moore:
As noted by JM, Open-with is not working on merge-ci.
Added more logging in last commit...

will-moore:
Updated and fixed the open-with config. Now deploying as expected.

manics:
It's be good if we could consider how to make automated deployments easier to.

In https://github.com/ome/ansible-example-omero-public-user/pull/5 I concluded the easiest way to get the script in place without hardcoding a version was to copy it:
```yaml
    - name: Copy Figure_To_Pdf.py script
      become: true
      become_user: omero-server
      copy:
        dest:
          /opt/omero/server/OMERO.server/lib/scripts/omero/figure_scripts/Figure_To_Pdf.py
        remote_src: true
        src:
          /opt/omero/web/venv3/lib/python3.6/site-packages/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
```

For example, if `omero figure ...` could do one of the following without a login
- print out the location of the file
- print the file to stdout
- copy the file to a defined location `$OMERODIR/lib/scripts/omero/figure_scripts/`
It'd be possible to run `omero figure ...` unconditionally as part of the OMERO.server startup commands.

This doesn't need to block this PR, but I'm mentioning it in case it affects the CLI API which we can't break after release

will-moore:
@manics Copying the script could work when omero-web is installed on the same machine (same $OMERODIR) as the server. So I guess we could add that as another / alternative command e.g:

```
$ omero figure script upload.  # currently this is just $ omero figure script

$ omero figure script copy
```

Maybe we need to discuss this PR, figure commands API etc now that the proof of concept is working

snoopycrimecop:
Conflicting PR. Removed from build [OMERO-plugins-push#567](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-plugins-push/567/). See the [console output](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-plugins-push/567/consoleText) for more details.
Possible conflicts:
  - Upstream changes
    - .travis.yml

--conflicts

will-moore:
Build failing because this PR is not merging
https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-web/680/console

Merge attempt fails, although current status of this PR looks good?
https://github.com/snoopycrimecop/omero-figure/commit/4cc755877a1d1d8af6069fbb8ae34e14f26c500f

```
Repository: ome/omero-figure
Excluded PRs:
  - PR 405 bramalingam 'Django upgrade 2 2' (status: )
  - PR 366 will-moore 'Omero figure cli' (status: )
Already up-to-date.

Generated by OMERO-plugins-push#568 (https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-plugins-push/568/)
```

For now I'll exclude this and update OMERO-web job to do a regular install

--exclude

PR #376, opened 2020-04-17
Fix nudge save

### LINKED ISSUE #372: Bug: nudge panels doesn't enable Save button
Reported by Michael.
"If I select a strip of images and use an arrow key to move them then the save button doesn’t enable. I have to then move a panel with the mouse and un-do for the button to enable."

will-moore:
Fixes #372.

To test:
 - Save a figure with at least 2 panels, then refresh page - Save button will be disabled
 - Select one panel and use arrow keys to nudge it. Save should be enabled.
 - Repeat: select multi-panels and "Align-Left" or "Align-Top" - these should also enable the Save button.

### DIFF
--- a/src/js/models/figure_model.js
+++ b/src/js/models/figure_model.js
@@ -472,7 +472,7 @@
 
             selected.forEach(function(p){
                 pos = p.get(axis);
-                p.set(axis, pos + delta);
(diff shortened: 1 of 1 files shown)


pwalczysko:
Works as expected on merge-ci. Ready to merge fmpov.

PR #377, opened 2020-04-18
fix channel label mismatch

### LINKED ISSUE #364: Channel name undefined
If you select multiple images that have different channel names, the channel buttons are labelled with ```undefined```.
Should simply be a ```-``` or empty string etc.

cc @jburel

will-moore:
If multiple images with same number of channels are selected, the channel buttons previously showed "undefined" if the labels didn't match.
Now we show no text.

Fixes #364.

Before:
Screenshot 2020-04-18 at 06 51 42

After:
Screenshot 2020-04-18 at 06 51 09

### DIFF
--- a/src/js/views/channel_slider_view.js
+++ b/src/js/views/channel_slider_view.js
@@ -244,7 +244,7 @@ var ChannelSliderView = Backbone.View.extend({
                 var max = maxs.reduce(reduceFn(Math.max));
                 var color = colors.reduce(allEqualFn, colors[0]) ? colors[0] : 'ccc';
                 // allEqualFn for booleans will return undefined if not or equal
-                var label = labels.reduce(allEqualFn, labels[0]);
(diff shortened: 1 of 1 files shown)


jburel:
The ``blank button`` should be shown if the name of the channel does not match
if for example we have 2 images with both DAPI then DAPI should be used instead of a blank button

will-moore:
That should be how it's working already?

e.g. 2 image selected - First channel is named differently but the other channels have same name in both images:
Screenshot 2020-04-18 at 21 32 22

jburel:
I thought so, I was a bit confused by the description in the PR

jburel:
Screenshot 2020-04-22 at 13 25 45

jburel:
Screenshot 2020-04-22 at 13 29 16

jburel:
Looks good

PR #378, opened 2020-04-19
Timestamp format millisecs

### LINKED ISSUE #375: More time-stamp formatting options
See https://forum.image.sc/t/small-feature-request-for-omero-figure/36103

"During a recent demo about OMERO and OMERO.figure, it was pointed out to us that the time stamps in the labels of OMERO.figure miss a few options, such as Time (msec) or Time (min:sec).
Would it be possible to add those options, please?"

will-moore:
Fixes #375 

This adds supports for displaying timestamp labels in milliseconds or mins:secs.
Previously, all timestamps were loaded as *seconds* (timestamps were rounded when images were added to figures) and these integer values are stored in the figure JSON, so all existing figures can't have timestamps in milliseconds.

So, when older figures are open (VERSION 4) we now load un-rounded timestamps from the server to update those values in the figure JSON.
This allows us to choose "milliseconds" for timestamp labels (also "mins:secs") AND fixes the display of milliseconds for the label beside the T-slider.

To test:
 - Use https://merge-ci.openmicroscopy.org/web/figure/file/25179 (user-3) which was created with VERSION 4 (last release). In this figure, timestamps are rounded to seconds (integers).
 - Open this figure - Browser console will say "Transforming to VERSION 5" and "Load timestamps for images (3) [83221, 64166, 122915]".
 - Now timestamps should be available without rounding. Can see full values with File > Export as JSON.
 - Add new images - their timestamps should also not be rounded.
 - T-slider label should now show time as hh:mm:ss.ss e.g. `00:03:45.02`.
 - Add Time-labels, test using "milliseconds" and "mins:secs".
 - Save, reload. Check OK.
 - Export and check that all timestamps in PDF/TIFF match those in the web.

### DIFF
--- a/src/templates/labels_form_inner_template.html
+++ b/src/templates/labels_form_inner_template.html
@@ -26,7 +26,7 @@
                         <a href="#" data-label="[time-index]">Time (T-index)</a>
                     </li>
                     <li class="add_time_label" title="Label shows time in milliseconds">
-                        <a href="#" data-label="[time-millisecs]">Time (milli secs)</a>
+                        <a href="#" data-label="[time-millisecs]">Time (millisecs)</a>
                     </li>
                     <li class="add_time_label">
                         <a href="#" data-label="[time-secs]">Time (secs)</a>
--- a/src/templates/labels_form_inner_template.html
+++ b/src/templates/labels_form_inner_template.html
@@ -26,7 +26,7 @@
                         <a href="#" data-label="[time-index]">Time (T-index)</a>
                     </li>
                     <li class="add_time_label" title="Label shows time in milliseconds">
-                        <a href="#" data-label="[time-millisecs]">Time (millisecs)</a>
+                        <a href="#" data-label="[time-milliseconds]">Time (milliseconds)</a>
                     </li>
                     <li class="add_time_label">
                         <a href="#" data-label="[time-secs]">Time (secs)</a>
--- a/src/templates/labels_form_inner_template.html
+++ b/src/templates/labels_form_inner_template.html
@@ -23,13 +23,19 @@
                         <a href="#" data-label="[key-values]">Key-Value Pairs</a>
                     </li>
                     <li title="Add a label that shows the current T-index">
-                        <a href="#" data-label="[time-index]">Time (index)</a>
+                        <a href="#" data-label="[time-index]">Time (T-index)</a>
+                    </li>
+                    <li class="add_time_label" title="Label shows time in milliseconds">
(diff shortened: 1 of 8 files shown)


src/templates/labels_form_inner_template.html
```diff
@@ -23,14 +23,20 @@
                         <a href="#" data-label="[key-values]">Key-Value Pairs</a>
                     </li>
                     <li title="Add a label that shows the current T-index">
-                        <a href="#" data-label="[time-index]">Time (index)</a>
+                        <a href="#" data-label="[time-index]">Time (T-index)</a>
+                    </li>
+                    <li class="add_time_label" title="Label shows time in milliseconds">
+                        <a href="#" data-label="[time-millisecs]">Time (millisecs)</a>
```
sbesson, line 29: I might I have raised the issue on a previously PR before but I would strongly recommend against the usage of custom abbreviation. Any reason not to use either the SI term `milliseconds` or the SI abbreviation `ms`?

jburel: agree 
The recommendations should be used

jburel:
The slider shows ``00:00:10.71``, as a user, I will expect to see the same  in the label

will-moore:
@jburel Which label? What format did you choose for the label?

jburel:
[mins:secs]

will-moore:
Then the label should say ```00:11``` (mins:secs) rounded to nearest second.

jburel:
it is rounded to the nearest second 10.47 -> 10 My point was more that the time is displayed in the slider as 10.47 but the precision is lost when adding as a label

will-moore:
If the user chooses ```mins:secs``` format then they are probably showing a bunch of frames from a movie that lasts several minutes, so they don't need to see milliseconds.
That's what I see when I look at figures from papers.
Sometimes it's even less precision. e.g. ```5 mins, 10 mins, 15 mins etc.```. If the microscope didn't get around to acquiring the image until 5 mins 1.52 secs, they don't want to show that on the figure.

will-moore:
We could add a format for ```mins:secs.00``` (or something) if we also want to support that.
It might be useful for FRAP example where some frames are very fast to start with, but the movie is longer than a few milliseconds?

will-moore:
Fixed "millisecs" and "secs" to "milliseconds" and "seconds".
Menu options now look like this:

Screenshot 2020-05-01 at 14 56 10

And formatted on the figure like this (uses "ms" and "s" units):
Screenshot 2020-05-01 at 14 56 23

jburel:
Thanks
We should probably do a release before the community meeting

PR #381, opened 2020-05-10
Ignore hidden scalebars

### LINKED ISSUE #379: Bug: Scalebar length KeyError in export script
From https://www.openmicroscopy.org/qa2/qa/feedback/29256/

"I keep getting this error when trying to export a figure. I don't understand where the problem is coming from......"

```
File "./script", line 2327, in <module>
run_script()
File "./script", line 2313, in run_script
File "./script", line 2269, in export_figure
File "./script", line 975, in build_figure
self.add_info_page(panels_json)
File "./script", line 1716, in add_info_page
sb_length = p['scalebar']['length']
KeyError: 'length'
```

Fix script to handle this nicely (use default length? or don't show scalebar) and see if there's any obvious way this could have been set in the UI?

will-moore:
This fixes a minor bug, brought to light by an error from user at #379.
The error was a KeyError because 'length' was missing from the scalebar dict (don't know why). The error came from the "Info-page" code, where we add scalebars to the the info page, and I noticed that this code was not taking the visibility of the scalebar into account.

Now, we ignore the hidden scalebars on the info page AND access length with ```scalebar.get('length') to avoid KeyError.
Fixes #379 

To test:
 - Add scalebars of different lengths to 2 different panels
 - Hide 1 of the scalebars and export the PDF
 - The hidden scalebar should be ignored in the "Scalebar lenghts" of the page-info

### DIFF
--- a/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
+++ b/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
@@ -1742,8 +1742,8 @@ class FigureExport(object):
         for p in panels_json:
             iid = p['imageId']
             # list unique scalebar lengths
-            if 'scalebar' in p:
-                sb_length = p['scalebar']['length']
+            if 'scalebar' in p and p['scalebar'].get('show'):
+                sb_length = p['scalebar'].get('length')
                 symbol = u"\u00B5m"
                 if 'pixel_size_x_symbol' in p:
                     symbol = p['pixel_size_x_symbol']


jburel:
The scale bar is no longer in the info page.
Now the confusion starts, which scale corresponds to which image?
When all the scale bars are visible, scale order = image order in ``Figure contains the following images``.
Now user will not know with one corresponds to which image

will-moore:
That's kinda outwith the scope of this PR.
If you can think of a nice way to indicate that, then it *might* be worth adding, but I think it's quite an edge-case.

Usually you only have a single size of scalebar in a figure and the figure legend will say "Scale bars are 10 microns". If you have multiple different sizes of scalebar users tend to add a label to each scalebar so this info is is not really needed in the figure legend, but it doesn't do any harm.

jburel:
One option could have the scale bar length after the link of the image.

I don't think this is outside this PR, since now the user will be confused since one is hidden and have a different scale bar

will-moore:
This PR fixes a Bug: if I have 2 images, only 1 of them shows a scalebar of 10um (the other has scalebar 20 um but the scalebar is hidden), but the info-page says "Scalebar Lengths 10 um, 20 um". This is fixed so the info-page ignores the hidden scalebar and correctly says "Scalebar Lengths 10 um". That is less confusing than it was before.

jburel:
but the user does not know which one is the scale bar

jburel:
Discussed privately with @will-moore, changing the current layout of the PDF will be challenging. 
I am still not convinced of the need of the Scale bar feature in the PDF when we have the scale bars on the images (if not hidden)

This PR fixes the problem in the current implementation

PR #380, opened 2020-05-09
Update changelog for 4.2.1 release

will-moore:
I think this is enough to justify a release...

PR #382, opened 2020-05-10
Changelog 4.3.0

will-moore:
Was previously opened as #380


jburel:
#381 could be included if it gets the green light.

will-moore:
Added #381 .
https://github.com/ome/omero-figure/pull/383 is merged now so can test when deployed tomorrow. I could mention it in the changelog (but quite edge-case fix)

PR #383, opened 2020-05-10
Fix negative timestamps format

will-moore:
Noticed bug with negative timestamps using hh:mm:ss format or mm:ss format since we round the hrs and minutes *down* in the UI and export script.

To test, check that the timestamps are shown correctly on this figure (screenshot shows current bug) e.g. ```0:00:-1``` and ```-1:59:59```

Screenshot 2020-05-11 at 00 04 15

### DIFF
--- a/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
+++ b/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
@@ -1089,6 +1089,8 @@ class FigureExport(object):
     def get_time_label_text(self, delta_t, format):
         """ Gets the text for 'live' time-stamp labels """
         # format of "secs" by default
+        isNegative = delta_t < 0
+        delta_t = abs(delta_t)
         text = "%d s" % int(round(delta_t))
         if format == "milliseconds":
             text = "%s ms" % int(round(delta_t * 1000))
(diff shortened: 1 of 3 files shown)


jburel:
Something is out of synch in the display
Screenshot 2020-05-12 at 10 45 54

jburel:
Check the time values one is ``-00:00:value`` one is ``00:00:-value``

will-moore:
@jburel Sorry - didn't list this for standup today but ```00:00:-value``` is fixed in right panel.
One issue I noticed: If you have a negative time (e.g 100 milliseconds) and you show ```hh:mm:ss`` or ``hh:mm``, they will currently appear as ```-00:00:00```, which is technically correct (they are negative) but does that look too wrong?

jburel:
Right-hand panel now
Screenshot 2020-05-13 at 11 43 54

jburel:
Screenshot 2020-05-13 at 11 53 51
Having the "negative" zeros do not look as you say, in that case it should ``00:00``

jburel:
This looks better
Screenshot 2020-05-14 at 13 36 16

jburel:
Outside the scope of this PR but for long lasting timelapse ``[hrs]`` and ``[hrs:mins]`` will be useful

jburel:
Screenshot 2020-05-14 at 13 53 44

jburel:
RFE https://github.com/ome/omero-figure/issues/384

PR #389, opened 2020-05-17
update doc

jburel:
Update for figure.o.org

cc @will-moore

### DIFF
--- a/index.html
+++ b/index.html
@@ -117,8 +117,6 @@ title: OMERO.figure
         <a id="install"></a>
         <h1>Installing from PyPI</h1>
         <p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -11,7 +11,7 @@ <h2 style="margin:0px">
           <div style="text-align:right; font-size:50%">Read more...</div>
           </a>
         </h2>
-        <h2>Current Version: <strong>4.0.1</strong>
```
joshmoore, line 14: This should probably be included in .bumpversion.cfg then.

jburel: Proposal is to redirect figure.o.org to www.o.org to reduce the maintenance

PR #390, opened 2020-05-17
update doc

jburel:
Update for figure.o.org
cc @will-moore

### DIFF
--- a/index.html
+++ b/index.html
@@ -117,8 +117,6 @@ title: OMERO.figure
         <a id="install"></a>
         <h1>Installing from PyPI</h1>
         <p>
-          OMERO.figure version 4.3.0 requires OMERO.web <strong>5.6.0 or newer</strong>.
-          We assume that OMERO.web is already installed.
           <ul>
             <li>Install the app using <a href="https://pip.pypa.io/en/stable/">pip</a>:<br>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -192,14 +197,14 @@ <h3>Upgrade the app</h3>
             </li>
             <li>Install the app using <a href="https://pip.pypa.io/en/stable/">pip</a>:<br>
               <code style="width:650px">
-                $ pip install omero-figure
+                $ pip install -U omero-figure
```
will-moore, line 200: I think we can remove this whole section of "upgrade from a version of OMERO.figure prior to 2.0.1" since that is quite old now.

will-moore:
Thanks for the update.
Can probably remove the sentence "OMERO.figure version 2.x requires OMERO.web 5.2.6 or newer." It's still true but seems kinda dated.

jburel:
@will-moore I have just also noticed we still point to jcb viewer. We will need to sort that out
Maybe time to redirect figure.o.org to another page e.g. www.o.org

will-moore:
Yes, I think we can just redirect to https://www.openmicroscopy.org/omero/figure/ now and save ourselves the maintenance of the old site.

PR #396, opened 2020-07-07
Replace cgi.escape by html.escape.

gouttegd:
The escape function in the cgi module has been deprecated since
Python 3.2 [1] and has finally been removed in Python 3.8 [2],
resulting in the Figure_To_Pdf.py script failing to run under
Python 3.8.

This patch replaces the call to cgi.escape by html.escape, as
recommended by the Python developers.

[1] https://docs.python.org/3.7/library/cgi.html#cgi.escape
[2] https://docs.python.org/3.8/whatsnew/3.8.html#api-and-feature-removals

### DIFF
--- a/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
+++ b/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
@@ -20,7 +20,7 @@
 import logging
 import json
 import numpy
-import cgi
+import html
 
 from datetime import datetime
 import os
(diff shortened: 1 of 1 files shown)


will-moore:
Thanks - looks good.
cc @knabar 

NB: we haven't looked at supporting Python 3.8 yet (see https://docs.openmicroscopy.org/omero/5.6.1/sysadmins/version-requirements.html#python) but happy to receive any help in that direction.

gouttegd:
For what it’s worth, I’ve been running OMERO under Python 3.8 for a little bit more than one month, and the problem with the Figure_To_Pdf.py script is the only issue I’ve run into so far.

PR #405, opened 2020-11-05
Django upgrade 2 2

bramalingam:
Linked to the omero-web Django upgrade PR: https://github.com/ome/omero-web/pull/200

reference comments on the requirement for this PR: https://github.com/ome/omero-web/pull/200#issuecomment-712117842

### DIFF
--- a/omero_figure/urls.py
+++ b/omero_figure/urls.py
@@ -18,60 +18,60 @@
 
 from omeroweb.webgateway import views as webgateway_views
 from . import views
-from django.conf.urls import url
+from django.urls import re_path, path
 
 
 urlpatterns = [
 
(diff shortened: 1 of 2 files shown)


bramalingam:
@will-moore : looks like we don't have black formatting in these plugin repositories, tried using black formatting but it did not match the formatting requirements, thus manually trying to fix the flake8 issues.

bramalingam:
And looks like dependencies are managed via omero-test-infra. Will review that repository to understand how dependencies are managed for the integration testing.

bramalingam:
Guess Django is being installed as a sub-dependency of omero-web, guess the open PR for omero-web might need to be integrated with this PR, please correct me if am wrong here.

PR #409, opened 2020-12-03
Draw rectangle fix

emilroz:
Not sure about the cause or how to reproduce as it's been reported to us via support channel. In some conditions it's possible to run into the following issue:

```

  File "./script", line 2361, in <module>

    run_script()

  File "./script", line 2347, in run_script

  File "./script", line 2303, in export_figure

  File "./script", line 977, in build_figure

    self.add_panels_to_page(panels_json, image_ids, page)

  File "./script", line 1803, in add_panels_to_page

    image, pil_img = self.draw_panel(panel, page, i)

  File "./script", line 1634, in draw_panel

    self.paste_image(pil_img, img_name, panel, page, dpi)

  File "./script", line 2055, in paste_image

    ShapeToPilExport(pil_img, panel, crop)

  File "./script", line 522, in __init__

    super(ShapeToPilExport, self).__init__(panel)

  File "./script", line 187, in __init__

    getattr(self, 'draw_%s' % s['type'].lower(), lambda s: None)(s)

  File "./script", line 641, in draw_rectangle

    stroke_width = scale_to_export_dpi(shape.get('strokeWidth', 2))

  File "./script", line 111, in scale_to_export_dpi

    return pixels * 300//72

TypeError: unsupported operand type(s) for //: 'str' and 'int'
```

A fixed proposed here address it directly in the Export script. Not sure if it's possible to fix it earlier at the Figure_JSON level.

### DIFF
--- a/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
+++ b/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
@@ -638,7 +638,7 @@ class ShapeToPilExport(ShapeExport):
         p.append(p[0])
         points = p
 
-        stroke_width = scale_to_export_dpi(shape.get('strokeWidth', 2))
+        stroke_width = scale_to_export_dpi(int(shape.get('strokeWidth', 2)))
         buffer = int(ceil(stroke_width) * 1.5)
 
         # if fill, draw filled polygon without outline, then add line later


omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
```diff
@@ -638,7 +638,7 @@ def draw_rectangle(self, shape):
         p.append(p[0])
         points = p
 
-        stroke_width = scale_to_export_dpi(shape.get('strokeWidth', 2))
+        stroke_width = scale_to_export_dpi(int(shape.get('strokeWidth', 2)))
```
joshmoore, line 641: It looks like [scale_to_export_dpi](https://github.com/ome/omero-figure/blob/80af88894ddac0f867dc06f827bafe59b13b8aae/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py#L106) can take floats. Does that lead to weirdness?

will-moore: Good catch @joshmoore. Yes, we do allow floats. The UI allows you to choose `0.25, 0.5, 0.75, 1, 2...`
I see that `scale_to_export_dpi(shape.get('strokeWidth', 2))` occurs in a few other places too, so they'll need the same fix too.
I should probably review `scale_to_export_dpi(pixels)` to do it's own casting of `pixels` to `float` but that might change the return type (int to float) in some cases so I'd want to check the impact of that.

jburel:
Replaced by #426

PR #410, opened 2020-12-08
Crop rotation fixes

### LINKED ISSUE #324: rotate and crop errors
From https://www.openmicroscopy.org/community/posting.php?mode=quote&f=4&p=20674
A user here at HMS reported to me that they were having problems with rotated and cropped figures. I was able to reproduce the problem. When you crop a figure that has been rotated the position of the cropping is not consistent with the output you get. The attached screenshot illustrates what is happening. The Preview pane shows what is intended. The resulting figure is offset to the lower left. The problem seems to be in the accuracy of the crop tool. If you reposition the crop frame only slightly, even though it appears to be in the wrong place, the correct crop is possible.

### LINKED ISSUE #408: ROI rectangle inset rotation
See https://forum.image.sc/t/omero-figure-box-annotations-rotated-incorrectly/45787

Copy the crop region from a rotated image and paste on it's "parent" image - but the new Rectangle isn't rotated.

Just playing with css tools, a fix can be simulated with a transform of negative rotation and transform-origin. See:

Screenshot 2020-11-26 at 10 27 43

But, we'd need to save this and handle it somehow when Editing the rectangle?

will-moore:
Fixes #408 - copy rotated crop region and paste to ROI.
Also fixes #324

Rectangle ROIs don't support rotation, but this fixes the issue above by creating a Polygon to simulate a rotated Rectangle.

To test:
 - Copy & Paste to duplicate a panel, then zoom in to crop, and rotate one of them.
 - Copy the Crop region and Paste it as a new Shape on the un-zoomed panel - The new Shape (Polygon) should match the crop region:

Screenshot 2020-12-08 at 21 55 42

 - Copy the rotated Crop region, select the uncropped (unrotated) Panel and click Paste to paste the rotated crop region. The rotation should also be applied to the panel so that both images show the same rotated viewport.
 - Crop the rotated panel using the "Crop" button. The 'initial' crop region should match the current viewport (see screenshot) and the updated crop region should be correctly applied to the panel.

Screenshot 2020-12-15 at 14 28 10

### DIFF
--- a/src/js/views/crop_modal_view.js
+++ b/src/js/views/crop_modal_view.js
@@ -140,8 +140,12 @@ var CropModalView = Backbone.View.extend({
                 y = parseInt($roi.attr('data-y'), 10),
                 width = parseInt($roi.attr('data-width'), 10),
                 height = parseInt($roi.attr('data-height'), 10),
+                rotation = $roi.attr('data-rotation') || 0,
                 theT = parseInt($roi.attr('data-theT'), 10),
                 theZ = parseInt($roi.attr('data-theZ'), 10);
+
+            // Rectangle ROIs have NO rotation. Copy of crop might have rotation
+            this.m.set('rotation', parseInt(rotation));
 
             this.m.set({'theT': theT, 'theZ': theZ});
 
@@ -164,8 +168,6 @@ var CropModalView = Backbone.View.extend({
             // var json = this.processForm();
             var self = this,
                 r = this.currentROI,
-                theZ = this.m.get('theZ'),
-                theT = this.m.get('theT'),
                 sel = this.model.getSelected(),
                 sameT = sel.allEqual('theT');
                 // sameZT = sel.allEqual('theT') && sel.allEqual('theT');
@@ -252,7 +254,7 @@ var CropModalView = Backbone.View.extend({
                         m.unset('shapes');
                     }
                     // 'save' to trigger 'unsaved': true
-                    m.save({'theZ': newZ, 'theT': newT});
+                    m.save({ 'theZ': newZ, 'theT': newT, 'rotation': self.m.get('rotation')});
                 });
             }
 
@@ -323,13 +325,13 @@ var CropModalView = Backbone.View.extend({
 
         showClipboardFigureRois: function() {
             // Show Rectangles from clipboard
-            var imageRects = [],
-                clipboardRects = [],
+            var clipboardRects = [],
                 clipboard = this.model.get('clipboard');
             if (clipboard && clipboard.CROP) {
                 roi = clipboard.CROP;
                 clipboardRects.push({
-                    x: roi.x, y: roi.y, width: roi.width, height: roi.height
+                    x: roi.x, y: roi.y, width: roi.width, height: roi.height,
+                    rotation: roi.rotation
                 });
             } else if (clipboard && clipboard.SHAPES) {
                 clipboard.SHAPES.forEach(function(roi){
@@ -487,6 +489,8 @@ var CropModalView = Backbone.View.extend({
                 left = -(zoom * rect.x);
                 rect.theT = rect.theT !== undefined ? rect.theT : origT;
                 rect.theZ = rect.theZ !== undefined ? rect.theZ : origZ;
+                let rotation = rect.rotation || 0;
+                let css = this.m._viewport_css(top, left, img_w, img_h, size, size, rotation);
 
                 var json = {
                     'msg': msg,
@@ -494,6 +498,7 @@ var CropModalView = Backbone.View.extend({
                     'rect': rect,
                     'w': div_w,
                     'h': div_h,
+                    'css': css,
                     'top': top,
                     'left': left,
                     'img_w': img_w,
--- a/src/js/views/crop_modal_view.js
+++ b/src/js/views/crop_modal_view.js
@@ -8,6 +8,16 @@ var CropModalView = Backbone.View.extend({
 
         model:FigureModel,
 
+        rotatePoint: function (x, y, cx, cy, rotation) {
+            // Get coordinates for point x, y rotate around cx, cy, by rotation degrees
+            let length = Math.sqrt(Math.pow((x - cx), 2) + Math.pow((y - cy), 2));
+            let rot = Math.atan2((y - cy), (x - cx));
+            rot = rot + (rotation * (Math.PI / 180));  // degrees to rad
+            let dx = Math.cos(rot) * length;
+            let dy = Math.sin(rot) * length;
+            return { x: cx + dx, y: cy + dy };
+        },
+
         initialize: function() {
 
             var self = this;
@@ -22,6 +32,16 @@ var CropModalView = Backbone.View.extend({
 
                 // get selected area...
                 var roi = self.m.getViewportAsRect();
+                var rotation = self.m.get('rotation');
+                if (rotation != 0) {
+                    var img_cx = self.m.get('orig_width') / 2;
+                    var img_cy = self.m.get('orig_height') / 2;
+                    var rect_cx = roi.x + (roi.width / 2);
+                    var rect_cy = roi.y + (roi.height / 2);
+                    var new_c = self.rotatePoint(rect_cx, rect_cy, img_cx, img_cy, rotation);
+                    roi.x = new_c.x - (roi.width / 2);
+                    roi.y = new_c.y - (roi.height / 2);
+                }
 
                 // Show as ROI *if* it isn't the whole image
                 if (roi.x !== 0 || roi.y !== 0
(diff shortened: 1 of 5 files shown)


pwalczysko:
> Copy the rotated Crop region, select the uncropped (unrotated) Panel and click Paste to paste the rotated crop region. The rotation should also be applied to the panel so that both images show the same rotated viewport.

Up till this point all worked. The above one I am not sure what to do with - where is the unrotated uncropped Panel ? I just used it in the previous step, and pasted the ROI on it, and it was rotated.... as I would expect.

will-moore:
> Copy the rotated Crop region, select the uncropped (unrotated) Panel and click Paste to paste the rotated crop region.

This means to use the "Paste" button at the bottom of the "Preview" Tab, to apply the same crop and rotation to the selected panel. This is NOT Pasting to create a new ROI as in previous steps.

pwalczysko:
> > Copy the rotated Crop region, select the uncropped (unrotated) Panel and click Paste to paste the rotated crop region.
> 
> This means to use the "Paste" button at the bottom of the "Preview" Tab, to apply the same crop and rotation to the selected panel. This is NOT Pasting to create a new ROI as in previous steps.

Thanks, got this one, and it works as expected.

pwalczysko:
> Crop the rotated panel using the "Crop" button. The 'initial' crop region should match the current viewport (see screenshot) and the updated crop region should be correctly applied to the panel.

Sorry, struggling with this one too. Using the Crop button is clear, but should I first rotate and then Crop ? This means I should not zoom first ? Should I copy and paste the ROI somewhere ?  If I do as told here, I am ending up with see screenshot below, how do I get the ROI displayed on the rotated image as shown on yours ?

Screenshot 2021-01-11 at 18 05 12

pwalczysko:
> Crop the rotated panel using the "Crop" button. The 'initial' crop region should match the current viewport (see screenshot) and the updated crop region should be correctly applied to the panel.

After clarification, I was able to verify the above workflow too, it works fine. 

Nevertheless, in this workflow, if the ROI does not come from the zoomed-in region but, is copied from another place in Figure (screenshot 1 below) or comes from OMERO (screenshot 2 below), this workflow still fails, as the copied or "from OMERO" ROIs are displayed unrotated.

Screenshot 1 (ROI copied from somewhere else in Figure and pasted (option "From Clipboard")

Screenshot 2021-01-12 at 11 17 17

Screenshot 2 (ROI from OMERO in Crop dialog, ROI should be rotated as the image is rotated)

Screenshot 2021-01-12 at 11 18 19

pwalczysko:
Would it be possible to fix also the Edit ROI panel view for rotated images ?

Workflow: Rotate an image in Figure. Click on "Labels" tab and "Edit" in the "ROI" row. Observe that the view in the Edit ROI dialog is unrotated, the image inside the panel on the canvas is rotated though.

Screenshot 2021-01-12 at 11 25 02

pwalczysko:
In summary, 3 RFEs : 2 RFEs regarding the crop dialog and rotation https://github.com/ome/omero-figure/pull/410#issuecomment-758591609 and one regarding the Edit ROI dialog https://github.com/ome/omero-figure/pull/410#issuecomment-758593581

will-moore:
@pwalczysko I'm afraid fixing the "Edit ROIs" dialog to handle rotated images isn't easy since we'd need drawing tools to allow you to draw rotated Rectangles which I don't have (hand-coded drawing tools).

But the other 2 RFEs above should be fixed:
 - Open the crop dialog, having either copied a Crop Region from another Panel (with or without rotation) OR having copied a Rectangle. Also nice to have Rectangle ROIs saved on the Image in OMERO, and/or Rectangle ROIs on the Figure panel (see screenshot)
 - Check that each of these thumbnails shows the correct region of the image.
 - Clicking either the Rectangle ROIs from OMERO or the Rectangle on the figure panel (top and bottom sections of thumbnails) will set the Panel rotation to zero. Clicking the "clipboard" thumbnail will apply a rotation IF you copied the crop region from a Panel that was rotated.
 - When you hit "OK", any rotation that got applied should be saved (so that the crop region you selected was correctly applied to the image).

Screenshot 2021-01-22 at 10 36 09

pwalczysko:
> @pwalczysko I'm afraid fixing the "Edit ROIs" dialog to handle rotated images isn't easy since we'd need drawing tools to allow you to draw rotated Rectangles which I don't have (hand-coded drawing tools).

I think this is fine. I would suggest a (couple of) short video(s) for the solution of the rotation and crop workflows would be very handy, as these workflows are very popular, but as the testing shows, not easy to get right on the first try.

Would you like to list for testing tomorrow with my name ?

pwalczysko:
> But the other 2 RFEs above should be fixed:

    Open the crop dialog, having either copied a Crop Region from another Panel (with or without rotation) OR having copied a Rectangle. Also nice to have Rectangle ROIs saved on the Image in OMERO, and/or Rectangle ROIs on the Figure panel (see screenshot)
    Check that each of these thumbnails shows the correct region of the image.
    Clicking either the Rectangle ROIs from OMERO or the Rectangle on the figure panel (top and bottom sections of thumbnails) will set the Panel rotation to zero. Clicking the "clipboard" thumbnail will apply a rotation IF you copied the crop region from a Panel that was rotated.
    When you hit "OK", any rotation that got applied should be saved (so that the crop region you selected was correctly applied to the image).

This works as expected. Ready to merge fmpov.

will-moore:
I made a demo movie for this PR:

https://user-images.githubusercontent.com/900055/106256644-731b4600-6213-11eb-90d1-e3c4173590b9.mp4

pwalczysko:
@will-moore great movie, thanks, these are precisely the workflows which were picked upon in the recent workshop again, and the difference of those two Paste buttons is nicely shown

jburel:
This could probably go to the official youtube channel otherwise it will be lost

will-moore:
Yes, can make the video more "public" when we release.
The last commit needs a test before merging: Copy the crop region from a non-rotated image (zero rotation) and paste as a crop region onto a rotated image. The rotation should be set to zero.

pwalczysko:
> Copy the crop region from a non-rotated image (zero rotation) and paste as a crop region onto a rotated image. The rotation should be set to zero.

Tested
When I use the Paste buttion from Labels pane, the rotation is not set to zero. When I use the Paste button from Preview pane, the rotation is not set to zero. I guess this is the expected behaviour ?

will-moore:
> When I use the Paste button from Preview pane, the rotation is not set to zero. I guess this is the expected behaviour ?

No. When I use the Paste button from Preview pane, the rotation should be set to zero.

pwalczysko:
> > When I use the Paste button from Preview pane, the rotation is not set to zero. I guess this is the expected behaviour ?
> 
> No. When I use the Paste button from Preview pane, the rotation should be set to zero.

Yes, the rotation on Pate is indeed set to zero. Sorry, typo (in my original report, there are two "no"s for both of the panes, in truth, I shoiuld have written "y" for the Preview pane, as this is what I saw.

Sorry

src/js/views/crop_modal_view.js
```diff
@@ -167,6 +179,18 @@ var CropModalView = Backbone.View.extend({
                 if (sameT) {
                     t = self.m.get('theT');
                 }
+
+                var rotation = self.m.get('rotation');
+                if (rotation != 0) {
```
jburel, line 184: could that section of code be turned into a function?
Same code above using ``roi`` instead of ``r``.

will-moore: Fixed in 88cafb4

pwalczysko:
Inside the Crop dialog, copied a Rrectangle ROI from another panel which was rotated so that it is in Clipboard now.

the desciption above says

> Clicking the "clipboard" thumbnail will apply a rotation IF you copied the crop region from a Panel that was rotated.

But when I click the Clipboard thumb it does not apply any rotation, instead, it unrotates the main panel in the Crop dialog and pastes the ROI correctly. Did this behaviour change (i.e. unitended change) or is it a misunderstanding ?

will-moore:
You copied a Rectangle ROI, but my text says "IF you copied the *crop region*..."
Rectangles can't have any rotation, but a crop region can.
The behaviour hasn't changed.

Does the thumbnail representation that you clicked look like the crop outline that is generated when you click it?

pwalczysko:
> You copied a Rectangle ROI, but my text says "IF you copied the _crop region_..."
> Rectangles can't have any rotation, but a crop region can.
> The behaviour hasn't changed.
> 
> Does the thumbnail representation that you clicked look like the crop outline that is generated when you click it?

thanks for explaining. 
No, the crop outline looks differently, the crop dialog is not showing the right thing now imho when Clipboard is selected.

Screenshot 2021-02-09 at 11 45 08

pwalczysko:
Compare the situation inside Crop dialog described in previous comment with the situation when I paste the crop region without using the Crop dialog on the same image.

Screenshot 2021-02-09 at 11 48 20

This behaviour seems right here, whereas the Crop dialog above https://github.com/ome/omero-figure/pull/410#issuecomment-775880924 shows the wrong thumb and when thumb is clicked, a wrong rotation and position is shown in main panel.

will-moore:
Sorry @pwalczysko I can't seem to reproduce the bug you're seeing.
Can you point me at an example figure and exact steps to reproduce? Thanks.

pwalczysko:
The image used is

https://merge-ci.openmicroscopy.org/web/webclient/?show=image-140172

The figure saved is

https://merge-ci.openmicroscopy.org/web/figure/file/70451

We clarified the workflow on a call - basically, taking the crop region from the zoomed in image, then selecting the non-zoomed non-rotated original image in the same figure and clicking on Crop button.  In the crop dialog, the thumb of the Clipboard is wrong, as well as when this wrong thumb is applied onto the image, it is not where it should be either.

pwalczysko:
The problem https://github.com/ome/omero-figure/pull/410#issuecomment-778292041 is fixed now.

Ready to merge fmpov.

PR #412, opened 2020-12-14
Fix scalebar units in PDF export

### LINKED ISSUE #406: Scalebar units in pdf wrong
As reported by the ns user and confirmed during our own testing:

- Create a figure
- Set the scalebar units to something else than micrometers, e.g. milimeters
- Export the figure and study the pdf
- On the second page of the pdf, note that the Text reads :  Scalebar Lengths: 10 μm
- this is not correct, as the Scalebar Lengths are in milimeters now, as can be verified on the first page of the pdf

cc @will-moore 
the-scalebar
wrong-units

will-moore:
Fixes #406 

To test:
 - Add scalebars with different units to different images
 - Export as PDF
 - Check that units are shown correctly in info page:

Screenshot 2020-12-14 at 14 36 44

### DIFF
--- a/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
+++ b/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
@@ -1749,8 +1749,9 @@ class FigureExport(object):
             if 'scalebar' in p and p['scalebar'].get('show'):
                 sb_length = p['scalebar'].get('length')
                 symbol = u"\u00B5m"
-                if 'pixel_size_x_symbol' in p:
-                    symbol = p['pixel_size_x_symbol']
+                sb_units = p['scalebar'].get('units')
+                if sb_units and sb_units in unit_symbols:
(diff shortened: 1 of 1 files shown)


jburel:
option is disabled. is the export script on merge-ci?

jburel:
tested using https://merge-ci.openmicroscopy.org/web/figure/file/64877
Screenshot 2021-01-07 at 16 16 41

The feature works, but the order of the scale does not corresponds to the order of the display
scalebar of the second image is first in the list

will-moore:
The images in a figure have no order (their position doesn't correspond to their order).
I collect the various scalebar lengths from all the images in the figure and only show each unique value, so it's likely that each scalebar applies to many images in different locations.
This export of scalebar lengths in the PDF was really needed before we had the ability to add labels to scalebars on the panels. Now that labels are supported (and will be the default with #413) the display of lengths in the PDF is not so important - and not worth spending too much time on improving (this is a bug-fix only).

jburel:
thanks for the clarification. Merging

Issue #407, opened 2020-11-11
BiorXiv paper on making figures


will-moore:
https://www.biorxiv.org/content/10.1101/2020.10.08.327718v1
Might have some useful info.

Creating Clear and Informative Image-based Figures for Scientific Publications

"We present detailed descriptions and visual examples to help scientists avoid common pitfalls when publishing images. Our recommendations address image magnification, scale information, insets, annotation, and color and may encourage discussion about quality standards for bioimage publishing."

will-moore:
Quality control feedback from workshop:
"User wants a warning on OMERO.figure - if you do not have a  scalebar on your image, pop up a warning when saving the figure: you do not have a scalebar, are you sure ?"

strambc:
It would be very useful to incorporate guidelines in Figure making.
QUAREP-LiMi (www.quarep.org) has just launched a Working Group on Publication Standards, which takes off from Elena's paper mentioned above, but will also include standards on Material and Methods.
It will be important to check the guidelines they produce and consider incorporating them into OMERO.Figure
Perhaps as warnings on basic requirements such as SCALE BAR, COLOR schema, INSET quality... would be useful.

will-moore:
The quarep checklist is also available at https://quarep-limi.github.io/WG12_checklists_for_image_publishing/image_publishing/1_image_format.html

One idea is to embed that checklist into OMERO.figure itself, including the description of each point and how to achieve it in OMERO.figure (if possible!). The checkboxes state could be saved in the figure json. That would be a big feature to add but not too tricky.

PR #413, opened 2020-12-14
scalebar show label by default

will-moore:
The paper https://www.biorxiv.org/content/10.1101/2020.10.08.327718v1.full
mentioned in https://github.com/ome/omero-figure/issues/407
has this advice about scalebars:

Annotate scale bar dimensions on the image: Stating the dimensions along with the scale bar allows readers to interpret the image more quickly.

This PR encourages users to add the label on the scale bar by showing the label by default.

To test:
 - Add a new Image panel
 - Add a scalebar - the label should be shown by default.
 - This won't change any existing figures where scale bars have been added to image panels already (even if the scale bar has subsequently been hidden) but if it's the first time that a scale bar has been added to a panel, the default 'show label' will apply.

### DIFF
--- a/src/js/views/scalebar_form_view.js
+++ b/src/js/views/scalebar_form_view.js
@@ -111,7 +111,7 @@ var ScalebarFormView = Backbone.View.extend({
     },
 
     render: function() {
-        var json = {show: false, show_label: false},
+        var json = {show: false, show_label: true},
             hidden = false,
             sb;
 


jburel:
* remove the label
*  hide scale
* tick label again
* scale bar is shown.
Is it the behaviour you were expected? since it is not mentioned in the description of this PR

The described workflow works

will-moore:
Yes, that behaviour is expected (not changed by this PR).
And the described workflow works. 👍

PR #414, opened 2020-12-14
Avoid duplicate labels from Key-Values

### LINKED ISSUE #395: Bug: Duplicate labels from Key-Value pairs
Feedback from OME 2020 workshop:
When creating labels from Key-Value pairs, it's possible to end up with 2 identical labels on the same panel. It's impossible to delete or edit just one of them.
Easy to avoid creating duplicates from KV pairs.
Harder to fix the handling of duplicate labels that are created by other means.

will-moore:
Fixes #395

To test:
 - Add duplicate Key-Value pairs to an Image
 - Choose that Key when adding labels to the image from key-value pairs.
 - Only 1 label should be created on each Panel for the duplicate Key-Value pairs.

### DIFF
--- a/src/js/views/labels_from_maps_modal.js
+++ b/src/js/views/labels_from_maps_modal.js
@@ -66,8 +66,10 @@ var LabelFromMapsModal = Backbone.View.extend({
                 prev[iid] = [];
             }
             t.values.forEach(function(kv) {
-                if (kv[0] === key) {
-                    prev[iid].push(kv[1]);
+                var value = kv[1];
+                // If key matches, add to values for the image (no duplicates)
+                if (kv[0] === key && prev[iid].indexOf(value) === -1) {
+                    prev[iid].push(value);
                 }
             });
             return prev;


jburel:
Discussed with @will-moore, it is currently not possible to add kvp in web using merge-ci.
This is currently broken

will-moore:
@jburel Thanks for finding that. Fixed in open PR: https://github.com/ome/omero-web/pull/225/commits/60dece6ba0fe6d69f5ee1d9d624661df0d7b1c9f

jburel:
using https://merge-ci.openmicroscopy.org/web/webclient/?show=image-83165 (user-1)
* created a duplicate kvp
* in figure, only one key listed 
Screenshot 2021-01-07 at 16 20 50
* I can add the same kvp multiple times (unexpected)
Screenshot 2021-01-07 at 16 23 20
* only one entry in the right-hand panel for the kvp (see screenshot)
* Click on red cross to remove, remove all the duplicated entries

will-moore:
Ah, that's a slightly different issue. In that case you have *chosen* to add the same label repeatedly, which will cause the same issue, regardless of whether it comes from KV pairs, or other metadata or is added manually.
I'm not sure that this is really a bug. Although you could consider the batch edit and delete behaviour to be unexpected. But that behaviour has always been the same an no-one has complained so I think it would need more discussion to change it and is outside the scope of #395 so I would prefer to just get the existing fix in for now.

PR #415, opened 2020-12-15
Handle save errors

### LINKED ISSUE #352: OptimisticLockException on Save
Need to handle this to let the user know the figure wasn't saved & please try again.

Seen today on nightshade.
```
Internal Server Error: /figure/save_web_figure/
  File "/opt/omero/web/venv/lib/python2.7/site-packages/django/core/handlers/base.py", line 132, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/omero/web/OMERO.web/lib/python/omeroweb/decorators.py", line 486, in wrapped
    retval = f(request, *args, **kwargs)
  File "/opt/omero/web/venv/lib/python2.7/site-packages/omero_figure/views.py", line 296, in save_web_figure
    update.saveAndReturnObject(fa._obj, conn.SERVICE_OPTS)
  File "/opt/omero/web/OMERO.web/lib/python/omero/gateway/__init__.py", line 4703, in __call__
    return self.handle_exception(e, *args, **kwargs)
  File "/opt/omero/web/OMERO.web/lib/python/omeroweb/webclient/webclient_gateway.py", line 2127, in handle_exception
    e, *args, **kwargs)
  File "/opt/omero/web/OMERO.web/lib/python/omero/gateway/__init__.py", line 4700, in __call__
    return self.f(*args, **kwargs)
  File "/opt/omero/web/OMERO.web/lib/python/omero_api_IUpdate_ice.py", line 163, in saveAndReturnObject
    return _M_omero.api.IUpdate._op_saveAndReturnObject.invoke(self, ((obj, ), _ctx))
OptimisticLockException: exception ::omero::OptimisticLockException
{
```

### LINKED ISSUE #365: Warn if save fails due to no permissions
## Public user
1. Open figure  as the public user
2. Create a new figure, add a public image
3. Click save. In the browser network console you'll see a 403 error but there's nothing in the UI to indicate it failed.

## Logged in user not in public group
1. Login as a user who is **not** a member of the public group
2. Browse the public group, since you should still be able to see it
3. Open figure
4. Create a new figure, add a public image
5. Click save. In the browser network console you'll see a 500 error but there's nothing in the UI to indicate it failed.

The latter is arguably a bigger problem. There is no expectation of the public user being able to save, however a user who is logged in but in the wrong group might still expect the save to work.

will-moore:
Fixes #365 and fixes #352 

This handles errors when saving figures.

EDIT - we now store the current figure JSON in local storage and direct the user to login page which redirects to /figure/recover/ page which will load the recovered figure.

Screenshot 2021-01-10 at 23 42 32

To test:
 - Create or Edit a figure
 - In another tab, log out of the webclient
 - Save the Figure - the above dialog should be shown
 - Clicking Reload button will open the webclient login page, then redirect you to `figure/recover/` page which should open the recovered file (unsaved). A dialog will ask if you want to clear the figure from local storage.
 - When you save the recovered figure, it will create a new file, to make sure we don't overwrite another version of the saved file.
 - File > Local Storage... gives you options to Clear local storage or to Recover a figure from local storage.

### DIFF
--- a/omero_figure/templates/figure/index.html
+++ b/omero_figure/templates/figure/index.html
@@ -47,12 +47,15 @@
         ACTIVITIES_JSON_URL = "{% url 'activities_json' %}",
         WEBGATEWAYINDEX = "{% url 'webgateway' %}",
         WEBINDEX_URL = "{% url 'webindex' %}",
+        WEBLOGIN_URL = "{% url 'weblogin' %}",
         ROIS_JSON_URL = "{% url 'api_rois' '0' %}",
         USER_FULL_NAME = "{{ userFullName }}",
         LENGTH_UNITS = {{ lengthUnits|safe }},
         IS_PUBLIC_USER = {% if isPublicUser %}true{% else %}false{% endif %},
         // Images larger than this are 'big' tiled images
         MAX_PLANE_SIZE = "{{ maxPlaneSize }}";
+
+    var LOCAL_STORAGE_RECOVERED_FIGURE = "recoveredFigure";
 
     $(document).ready(function() {
 
--- a/omero_figure/templates/figure/index.html
+++ b/omero_figure/templates/figure/index.html
@@ -663,6 +663,9 @@
                             <li class="save_as">
                                 <a href="#">Save a Copy...</a>
                             </li>
+                            <li class="local_storage" title="Open or Clear recovered data from local storage">
+                                <a href="#">Local storage...</a>
+                            </li>
                             <li class="export_json">
                                 <a href="#">Export as JSON...</a>
                             </li>
@@ -671,9 +674,6 @@
                             </li>
                             <li class="delete_figure">
                                 <a href="#">Delete</a>
-                            </li>
-                            <li class="local_storage" title="Open or Clear recovered data from local storage">
-                                <a href="#">Local storage...</a>
                             </li>
                             <li class="paper_setup">
                                 <a href="#">Paper Setup...</a>
--- a/omero_figure/templates/figure/index.html
+++ b/omero_figure/templates/figure/index.html
@@ -672,6 +672,9 @@
                             <li class="delete_figure">
                                 <a href="#">Delete</a>
                             </li>
+                            <li class="local_storage" title="Open or Clear recovered data from local storage">
+                                <a href="#">Local storage...</a>
+                            </li>
                             <li class="paper_setup">
                                 <a href="#">Paper Setup...</a>
                             </li>
--- a/omero_figure/urls.py
+++ b/omero_figure/urls.py
@@ -26,6 +26,7 @@ urlpatterns = [
     # index 'home page' of the figure app
     url(r'^$', views.index, name='figure_index'),
     url(r'^new/$', views.index, name='new_figure'),
+    url(r'^recover/$', views.index, name='recover_figure'),
     url(r'^open/$', views.index, name='open_figure'),
     url(r'^file/(?P<file_id>[0-9]+)/$', views.index, name='load_figure'),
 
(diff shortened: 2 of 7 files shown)


jburel:
I ran the workflow described in the PR
* Dialog should probably indicate to login before attempting the "save from json" step.
* Export as Json only shows a dialog and the content needs to be cut and paste. The message could be improved so the user does not look for a file on the local machine
* I logged in again and attempted to create the figure from the json using "Import from JSON"
* I copied the content from the previous step, clicked ok, nothing happens the "Import from JSON" dialog does nothing and stays on screen

will-moore:
@jburel I used a different strategy (auto-saving to localstorage and then recovering it after login) to make it easier for users and safer. See edited description above.

jburel:
* open a figure
* make a small change
* disconnect in web
* Click save
* Click on relaunch button
* Enter the login credentials
* No figure display at https://merge-ci.openmicroscopy.org/web/figure/recover/
Screenshot 2021-01-07 at 16 28 21
i.e. nothing is recovered

Also checking the connection should happen when for example the user click on export PDF otherwise it will spin forever and the user won't know that he/she is disconnected.

will-moore:
@jburel Sorry - that's a bug from another PR. Should be fixed by https://github.com/ome/omero-figure/pull/416/commits/2c3d787a9486960c53540c7fb88e0451746825b6

jburel:
This time, recovering from local storage works.
Ideally the ``Recover Figure`` button should be greyed out when there is no figure to recover.

jburel:
Maybe I am missing something 
When I recover the image from the Local Storage, the message indicates that the figure has been removed from the local storage. After that I will assume the Local Storage will be empty (I only have one figure in) but it is not the case, I can still recover the figure again and again.

Also the message when clicking on save could be improve
Title could be save: Error 403
and remove Error 403 in the core of the dialog. It does not give much since I cannot see the error.

Screenshot 2021-01-19 at 15 44 53

will-moore:
When you recover from local storage, it doesn't say that the figure has been removed from local storage:

Screenshot 2021-01-19 at 16 51 40

I don't automatically delete from local storage because the user may still not have saved it. It's better that the user explicitly deletes it when they want to. I don't think it's a problem to allow the user to recover the figure more than once.

I'll look at improving the error...

jburel:
@will-moore I read "removed" instead of "recovered" leading to the comment. In that case, that's fine.
Only one dialog to fix in that case

will-moore:
Dialog updated:

Screenshot 2021-01-20 at 23 06 17

jburel:
typo noticed when clicking on local storage after clear it.
Screenshot 2021-01-21 at 17 33 20

I assume you mean ``currently``

jburel:
Previous comment 
Will you implement the same for Save to PDF?
it spins if disconnected

will-moore:
@jburel Thanks - typo fixed.
I'm not sure what to do for Exporting figures. I'd want to distinguish between connection errors and other script export errors. I guess I would handle connection failures in a similar way, but maybe just pop-up a login dialog.
I was a bit more paranoid for handling Save errors since the app gets into a more messy state (needs a refresh) and there is the concern about losing data. Export failure is less of a worry since you can probably re-login and the Export or Save as normal.
But I'd prefer not to do it in this PR. It's less urgent (not worried about data loss) and this PR is already quite long, and conflicting with [#416](https://github.com/ome/omero-figure/pull/416#issuecomment-758319969).
So if this is good to merge, I can fix that conflict etc.
I'll created an issue for handling figure Export failures: https://github.com/ome/omero-figure/issues/418

jburel:
Agree that we can do it another PR, it certainly does not seem to put the app in a nasty state,

jburel (approved):

PR #416, opened 2020-12-17
Show group info

### LINKED ISSUE #284: figure group permissions are not obvious
Figures belong to both a user and a group.  The first is pretty obvious and when searching for existing figures, one can filter by user.  However, figure group permissions are never referred on the GUI.  Since a user can belong to multiple groups, it's not obvious to the user which group his figures are.

The problem we just had were two users that belong to different primary groups and a same secondary group.  They wanted to share figures with images from this secondary group.  Because they were not aware of group permissions for figures, some of them ended up on their primary group and so were invisible to the other user.

The GUI should make some reference to the group a figure is in.  I would guess a 'Move to group' button on the 'File' menu may also be useful.

will-moore:
Fixes #284 to show which groups each Figure is in and allow you to chgrp.

To test:

 - File > Open list now includes groups and allows you to filter by Group just the same as Owner.
 - File > Move Figure to Group... shows a chgrp dialog where you can pick a group to move to. Chgrp should be "instant" (we wait for a response instead of polling for completion) since it's not expected to take long. 
 - This dialog also shows you which group the images in the figure belong to, with links to the images. This should help to ensure you keep the images and figure in same group.

Screenshot 2020-12-17 at 12 09 52

Screenshot 2020-12-17 at 14 26 52

### DIFF
--- a/omero_figure/templates/figure/index.html
+++ b/omero_figure/templates/figure/index.html
@@ -44,11 +44,13 @@
         LIST_WEBFIGURES_URL = "{% url 'list_web_figures' %}",
         DELETE_WEBFIGURE_URL = "{% url 'delete_web_figure' %}",
         MAKE_WEBFIGURE_URL = "{% url 'make_web_figure' %}",
+        API_BASE_URL_V0 = "{% url 'api_base' '0' %}",
         ACTIVITIES_JSON_URL = "{% url 'activities_json' %}",
         WEBGATEWAYINDEX = "{% url 'webgateway' %}",
         WEBINDEX_URL = "{% url 'webindex' %}",
         ROIS_JSON_URL = "{% url 'api_rois' '0' %}",
         USER_FULL_NAME = "{{ userFullName }}",
+        USER_ID = {{ userId }},
         LENGTH_UNITS = {{ lengthUnits|safe }},
         IS_PUBLIC_USER = {% if isPublicUser %}true{% else %}false{% endif %},
         // Images larger than this are 'big' tiled images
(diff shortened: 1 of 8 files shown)


pwalczysko:
Simple workflow (first two bullet points in the description) works as described. The third bullet point seems to work too, but

https://merge-ci.openmicroscopy.org/web/figure/file/25179 seems to indicate that the figure has 3 images which belong to the same group. When the "3 images" link is clicked, in the webclient, only 2 images are selected. What am I missing in this particular case ?

pwalczysko:
It would be good to be more verbose on the message which indicates in which groups are the images. The implicit message should be made more obvious by stating something like "Note that only a user with permissions to see the images included in the figure will be able to see all the panels inside this figure". Possibly this could be shortened, but the assumption that the user will know the above is not accurrate for many users.

snoopycrimecop:
Conflicting PR. Removed from build [OMERO-plugins-push#639](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-plugins-push/639/). See the [console output](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-plugins-push/639/consoleText) for more details.
Possible conflicts:
  - PR #415 will-moore 'Handle save errors'
    - omero_figure/templates/figure/index.html
    - src/js/views/figure_view.js

--conflicts

snoopycrimecop:
Conflicting PR. Removed from build [OMERO-plugins-push#640](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-plugins-push/640/). See the [console output](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-plugins-push/640/consoleText) for more details.
Possible conflicts:
  - PR #415 will-moore 'Handle save errors'
    - src/js/views/figure_view.js

--conflicts

pwalczysko:
Not sure what changed here since last testing. 
Would still go for the more verbose mesage https://github.com/ome/omero-figure/pull/416#issuecomment-756938811
The problme described in https://github.com/ome/omero-figure/pull/416#issuecomment-756927852 is not caused by this PR - interestingly, a link to 3 images when pasted into new browser window will highlight only 2 of those images in the webclient (the 2 ones from the same dataset) and the third one is ignored (not highlighted), see https://merge-ci.openmicroscopy.org/web/webclient/?show=image-64166|image-83221|image-122915

will-moore:
@pwalczysko  Apologies, I missed your previous comments before rebase etc.
I added more info to the dialog (see below).
For the link to the images, I just wanted a light-weight way that a user could track down the images that were in the "wrong" group, expecting that to be a small number. Eg. see screenshot. I could add a separate link to every image in the figure, but how to do that without it becoming a very bloated dialog (lots of image names etc)? I guess I could use IDs as text to link to individual images? But that could still end up with 20+ links, all to images in the same group you I don't need to find anyway. Or allow user to expand a section of the dialog, but that just takes more time, coding and testing.

Screenshot 2021-01-25 at 13 00 09

dominikl:
Looks good to me 👍 I think too, that it's better not to overload the dialog.

pwalczysko:
> I think too, that it's better not to overload the dialog.

Yes, this is fine imho, the point is not to put all the links into the dialog, but rather, to explain the concept of the Figure in Group and Images in (another) Group. I think the present state is sufficient.

jburel:
Merging so we can move forward with #421

PR #417, opened 2020-12-17
Jquery included

will-moore:
This PR updates OMERO.figure to host jquery.js, jquery-ui.js and jquery-ui.css instead of loading them from `omero-web` since they are all updated in https://github.com/ome/omero-web/pull/180 and `bootstrap 3.0.0` is not compatible with latest `jquery`. We load the exact same versions of jquery and jquery-ui, so there shouldn't be any issues. However, I haven't included all the jquery-ui.css since most of it shouldn't be needed.

According to https://github.com/twbs/bootstrap/issues/16834#issuecomment-251996660 "Bootstrap 3.3.7 now supports jQuery 3+." but this would require a lot of upgrade work on bootstrap in OMERO.figure, so this is a shorter-term fix to allow `omero-web` to upgrade jquery.

To test:
 - App should still be working. No JavaScript issues expected - same libraries are loaded.
 - No css issues (e.g. with jQuery UI components such as sliders or tabs)


will-moore:
This has been in the merge-ci deployment for a while, with no issues noticed.
Now excluding https://github.com/ome/omero-web/pull/180, to verify that OMERO.figure with this PR can be released independently of the `omero-web` jQuery update....

will-moore:
I think that figure should permanently host its own copy of jquery, just as it does with all other dependencies.
This is really the ending of the "exceptional" behaviour of sharing jquery with `omero-web`.
This behaviour hasn't been documented before since it's really internal behaviour and doesn't affect users or even developers.
 And now that is ending, so I don't think it will be of much interest to document?

jburel:
I agree that being "web independent" is preferable.
We should, at least, create an issue indicating that an upgrade of jquery is required for figure since as mentioned in the description `` this is a shorter-term fix...``

will-moore:
Issue created at https://github.com/ome/omero-figure/issues/423

PR #421, opened 2021-01-26
Handle save errors2

will-moore:
This is a follow-up to #415. cc @jburel 

NB: this is on top of #416 (needed the new code from that PR), so that should be merged first.

To test:
 - When a figure is recovered, the figure is cleared from local storage
 - We use 'sessionStorage' for recovering figures. This is cleared when you quit the browser
 - Figure is saved in a per-user storage.

### DIFF
--- a/omero_figure/templates/figure/index.html
+++ b/omero_figure/templates/figure/index.html
@@ -57,7 +57,7 @@
         // Images larger than this are 'big' tiled images
         MAX_PLANE_SIZE = "{{ maxPlaneSize }}";
 
-    var LOCAL_STORAGE_RECOVERED_FIGURE = "recoveredFigure";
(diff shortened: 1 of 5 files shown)


jburel:
@will-moore could you sort this one so it is easier to review without the commits from #416?

will-moore:
Rebased and pushed...

jburel:
Test 1:
-----
* Logged as user-1, opened a figure and made a change
* Disconnect
* Logged as user-2, go back to the figure tab and click Save
* Message indicating that an error occurred and that the figure is the local storage popped up
* Login page popped up, entered the credentials of user-2. 
* No figure displayed as expected

jburel:
The workflow using the same user works as expected

jburel:
Local storage is cleared when the browser is closed.

PR #422, opened 2021-01-29
Rois pagination

### LINKED ISSUE #334: Handle >1k ROIs
see https://forum.image.sc/t/omero-iviewer-cant-display-more-then-500-rois/27830

> I have images in OMERO with many ROIs. In the case of my specific example image its 1032 ROIs. When I open this image in OMERO-iviewer (0.6.0), only the first 500 ROIs are there. The same is true when opening the image in OMERO-figure (v4.0.2).

will-moore:
Fixes #334 

Adds support for loading ROIs by page in the main `Edit ROIs` dialog, as well as the `Crop` dialog.

To test ROIs dialog...

 - Find an image that has lots of ROIs - (over 500)
 - Open the `Edit ROIs` dialog and load ROIs from OMERO
 - The first page of 500 ROIs should load (as before) but if you scroll to the bottom, there is some info on how many you've loaded and a "Load More..." button.
 - Clicking the button should load more. When there are no more to load, the button should be hidden. ROIs should be added when you click `Add` button for a ROI.

Screenshot 2021-01-31 at 23 43 51

To test Crop dialog...

 - Find an image that has lots of ROIs - (over 200) including Rectangles
 - Open the Crop dialog. The first 200 ROIs are loaded and any Rectangles are shown as thumbnails for picking a crop region (as before).
 - Scrolling to the bottom shows info on how many ROIs loaded and a `Load More...` button. This should load more and the be hidden when no more to load.

Screenshot 2021-01-30 at 22 53 16

### DIFF
--- a/omero_figure/views.py
+++ b/omero_figure/views.py
@@ -580,10 +580,12 @@ def roi_rectangles(request, image_id, conn=None, **kwargs):
     and shapes.class = Rectangle"""
     params = omero.sys.ParametersI()
     params.addLong('image_id', image_id)
-    result = conn.getQueryService().projection(count_query, params, conn.SERVICE_OPTS)
-    totalCount = result[0][0].val
+    result = conn.getQueryService().projection(count_query, params,
+                                               conn.SERVICE_OPTS)
+    total_count = result[0][0].val
 
-    return JsonResponse({'data': json_data, 'meta': {'totalCount': totalCount}})
+    return JsonResponse({'data': json_data,
+                         'meta': {'totalCount': total_count}})
 
 
 @login_required()
--- a/omero_figure/views.py
+++ b/omero_figure/views.py
@@ -35,6 +35,7 @@ from omero.model import LengthI
 from omero.model.enums import UnitsLength
 from omero.cmd import ERR, OK
 import omero
+from omero_marshal import get_encoder
 
 from io import BytesIO
 
@@ -541,6 +542,48 @@ def unit_conversion(request, value, from_unit, to_unit, conn=None, **kwargs):
            'symbol': to_value.getSymbol()}
 
     return HttpResponse(json.dumps(rsp), content_type='application/json')
+
+
+@login_required()
+def roi_rectangles(request, image_id, conn=None, **kwargs):
+    """
+    Load ROIs that have Rectangles and marshal with omero_marshal
+
+    Returns similar JSON to /api/ rois, with meta.totalCount
+    Supports pagination with ?offset and ?limit
+    """
+
+    params = omero.sys.ParametersI()
+    params.addLong('image_id', image_id)
+    limit = request.GET.get('limit')
+    offset = request.GET.get('offset', 0)
+    if limit is not None:
+        params.page(int(offset), int(limit))
+    query = """select roi from Roi roi join fetch
+    roi.details.owner as owner join fetch roi.details.creationEvent
+    left outer join fetch roi.shapes as shapes
+    where roi.image.id = :image_id
+    and shapes.class = Rectangle
+    order by roi.id"""
+
+    rois = conn.getQueryService().findAllByQuery(
+        query, params, conn.SERVICE_OPTS)
+
+    json_data = []
+    for roi in rois:
+        encoder = get_encoder(roi.__class__)
+        json_data.append(encoder.encode(roi))
+
+    count_query = """select count(distinct roi) from Roi roi
+    left outer join roi.shapes as shapes
+    where roi.image.id = :image_id
+    and shapes.class = Rectangle"""
+    params = omero.sys.ParametersI()
+    params.addLong('image_id', image_id)
+    result = conn.getQueryService().projection(count_query, params, conn.SERVICE_OPTS)
+    totalCount = result[0][0].val
+
+    return JsonResponse({'data': json_data, 'meta': {'totalCount': totalCount}})
 
 
 @login_required()
--- a/omero_figure/templates/figure/index.html
+++ b/omero_figure/templates/figure/index.html
@@ -309,7 +309,7 @@
                                 <!-- ROIs loaded here... -->
                                 </tbody>
                             </table>
-                            <button style="float: right" class="btn btn-sm btn-default loadRoiRects">Load ROIs</button>
+                            <button style="float: right" class="btn btn-sm btn-default loadRoiRects">Load More...</button>
                             <!-- Additional message/status -->
                             <p style="color: #999; margin: 10px" id='cropRoiMessage'></p>
                         </div>
--- a/omero_figure/templates/figure/index.html
+++ b/omero_figure/templates/figure/index.html
@@ -309,8 +309,9 @@
                                 <!-- ROIs loaded here... -->
                                 </tbody>
                             </table>
-                            <!-- Additional message depends on whether we have any ROIs or not -->
-                            <p id='cropRoiMessage'></p>
+                            <button style="float: right" class="btn btn-sm btn-default loadRoiRects">Load ROIs</button>
+                            <!-- Additional message/status -->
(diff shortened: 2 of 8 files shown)


pwalczysko:
Works as described in the PR header. Tested on multi-z image with > 2000 ROIs on different z planes. https://merge-ci.openmicroscopy.org/web/webclient/?show=image-140172 (user-3)

Couple of RFEs, which might or might not be dealt with in this PR, I think up to @will-moore 

1. Crop dialog rectangles:  One has load them all to find out if there were any rectangles already (there were none, and I have drawn some in iviewer).These new rectangles were added to the bottom of the ROI list, and loaded only when I clicked 10 times on the Load ROIs button. Once I added a rectangle, it worked, and the crop was fine. But then I realized this is not an optimal rectangle, and I had to go back to the Crop dialog. All the ROIs were unloaded - so I had to again click 10 times to load the rectangles and select another rectangle. - suggestions: When I am in Figure, could the ROIs stay loaded once I loaded them into the Crop dialog ? Could there maybe be a function saying "there are no Rectangles here" instead of me loading all the ROIs just to find out there are no rectangles ?
2. Edit ROIs dialog: The preview is actually much less forgiving that the Crop dialog one. The ROIs are viewable only when hovered over in the list. There is no way to order the loaded ROIs according to z, so a mixture of the ROIs on different z planes in the same place of the table is the result. There is a pending RFE about Add all ROIs button, but I would think also Add all ROIs from z plane would be most useful too.


-

will-moore:
Thanks:

 - 1. We don't have any way to load *ONLY* Rectangle ROIs from OMERO using the JSON API (which is what I'm currently using). I could write a custom 'Rectangle loader' for figure, which would provide a nicer experience and might not be very much code. I'll look into it...
 - 2. I originally was going to add a preview for every Shape in the Edit ROIs dialog, but this is a lot of work to support previews for all types of Shape. Also, it would potentially load a lot of pixel data to show these. I guess we could immediately show them ALL on the image, but we'd need a way to distinguish those that are actually added to the image (and can be edited) and those that merely *could* be added from OMERO.

will-moore:
Now we load only ROIs with Rectangles for the crop dialog. That should fix RFE 1) above.

pwalczysko:
> Now we load only ROIs with Rectangles for the crop dialog. That should fix RFE 1) above.

@will-moore thank you , the RFE 1) above fix works as expected, it seems that it even nicely keeps the loaded rectangles there when repeated Crop dialog usage is done.

I have one more RFE Re: teh ROI Edit dialog. The info about the ROIs to be loaded could be on the top, not on the bottom (the bottom is actually not seen at first, only when you scroll there). This would spare the user to wonder "where are my ROIs ?" because they did not load them all as they do not know there are not all loaded by default.

Thus
This bit

Screenshot 2021-02-03 at 11 14 16

Could be here ?
Screensh![Screenshot 2021-02-03 at 11 14 16ot 2021-02-03 at 11 14 10](https://user-images.githubusercontent.com/2478303/106739737-603cb300-6611-11eb-82da-53db7682650e.png)

pwalczysko:
> I guess we could immediately show them ALL on the image, but we'd need a way to distinguish those that are actually added to the image (and can be edited) and those that merely could be added from OMERO.

I think so, but probably something for a different PR.. ?

omero_figure/views.py
```diff
@@ -543,6 +544,50 @@ def unit_conversion(request, value, from_unit, to_unit, conn=None, **kwargs):
     return HttpResponse(json.dumps(rsp), content_type='application/json')
 
 
+@login_required()
+def roi_rectangles(request, image_id, conn=None, **kwargs):
```
jburel, line 548: is not something that should be pushed to the API itself with something like ?type=...

will-moore: Yes that would be great, but that won't likely happen in the timeframe of this PR.

Support for `/api/rois/?class=Rectangle` could work (uses same terminology as the HQL). But we'd need a custom mapping for that URL to say "class applies to Shapes", whereas e.g. `/api/annotations/?class=TagAnnotation` the class would apply to the annotations themselves. Writing query-builder code that doesn't use custom mapping might use a notation like `/api/rois/?shapes.class=Rectangle`.

I can imagine that work taking a while, and I wouldn't want that to hold up this PR. But we can certainly revert to using the API once it's supported there.

jburel: we have explored in the past in "clients" possible API call.
The key point is to make sure we go back to it when the API call is implemented. 
Could you create an issue for the API call addition and figure update

will-moore: Issue created at https://github.com/ome/omero-web/issues/258

will-moore:
@pwalczysko Re: https://github.com/ome/omero-figure/pull/422#issuecomment-772433994
I don't think you'd be disappointed that it hadn't loaded all your ROIs, since you wouldn't know that some were missing until you scroll to the bottom of the list to check that they're not there - and then you'd see "Load more..." ?

To me, it makes sense for a simple "Load more..." to go at the end of the list like:

 - item
 - item
 - item
 - `Load more...`

Whereas for more complex pagination controls it makes sense to put at the top of the list:

`Prev page` | `Next page` | Page `1` `2` `3`
Showing Page 1 / 3:
 - item
 - item
 - item

I'll try and see how it looks...

pwalczysko:
> since you wouldn't know that some were missing 

You very possibly know your ROIs from iviewer or another viewer, even very probably I would say ? I guress you come to Figure with at least a partially pre-concieved idea about which ROIs to add and why, this would make sense to me...

But as a whole, I definitely do not see the RFE as a blocker, if too difficult to implement or looks bad, then we can leave it.

will-moore:
Yes, you know your ROIs, but how would you know that they haven't all loaded when you can only see:

Screenshot 2021-02-04 at 22 56 48

You have to scroll down to find your ROIs, then when you get to the bottom you'll see the `Load More...` button.
Moving the controls to the top is certainly a bit more work, since this involves a different UI and a bit of a redesign etc.

pwalczysko:
> Moving the controls to the top is certainly a bit more work, since this involves a different UI and a bit of a redesign etc.

aha, thanks. I think we should leave it then as it is.

will-moore:
Hmm - Actually, I hate to think of a case where I have thousands of ROIs, I add one more in iviewer, then I want to find it in figure, I'd want to jump to the *last* page, which could be painful. So I'm just going to do pagination - won't take too long...

will-moore:
Done. To test:
 - Try with image that has fewer than 500 ROIs, more than 500 ROIs and none:

Screenshot 2021-02-05 at 15 32 41

In the screenshot I temporarily limited the page size to 100 since the image didn't have > 500 ROIs.

pwalczysko:
Tested the last commit (pagination) with 

   - more than 2000 ROIs images
   - 3 and 20 ROIs images
   - no ROI image
 
Worked as expected.
Couple of minor RFEs in order of improtance:

1. Could the dropdown manu with the page list show what page I am corrently on ? This could be done with a tick mark, as is usual in other apps, see screenshot 1 below.
2. What is "Export Options" as tooltip on the dropdown menu next to "Next" button ? Does not sound very helpful. Maybe replace with "Select page" see screenshot 2 below.
3. If there is only one ROI page, the widget with Next and Prev and dropdown should not be shown at all.

1.
Screenshot 2021-02-08 at 11 46 32

2.
Screenshot 2021-02-08 at 11 37 18

will-moore:
That should address those 3 issues:
 - show current page on drop-down list
 - Fix tooltip (copy and paste error)
 - Don't show pagination controls unless needed (> 1 page)

pwalczysko:
> That should address those 3 issues:
> 
>     * show current page on drop-down list
> 
>     * Fix tooltip (copy and paste error)
> 
>     * Don't show pagination controls unless needed (> 1 page)

Can confirm the fix, nice behaviour now. Ready to merge fmpov.

PR #430, opened 2021-04-19
Add All button for ROIs table

### LINKED ISSUE #335: "Add all" ROIs option
https://forum.image.sc/t/omero-iviewer-cant-display-more-then-500-rois/27830

> Is there an option to “add all” ROIs to the image in the figure?

will-moore:
Fixes #335 

This provides an "Add All Shapes" button to add all the shapes that are on the current Z/T plane (or shapes with Z/T unset).
It ignores shapes that are outside the current viewport.

To test:

 - Need an image with various ROIs. NB: I needed this feature with image from idr0079: https://merge-ci.openmicroscopy.org/web/webclient/?show=image-187415 which I want to use for OME2021 figure workshop.
 - Test that "Add All Shapes" button adds all that are on the current Z/T index (as shown in top-left of the Image in the ROI dialog). Any that are outside of the current viewport are silently ignored (try zooming in to a region of the image in the Preview panel before adding ROIs)
 - In contrast, clicking the "Add" button on a particular Shape row should still show a message IF the shape is outside the viewport and is not added.

Screenshot 2021-04-19 at 14 40 25

### DIFF
--- a/docs/demo/index.html
+++ b/docs/demo/index.html
@@ -717,7 +717,7 @@ redirect_to:
                     <button id="pdf_inprogress" style="display:none" class="btn btn-primary btn-sm"
                             title="Figure generation in progress..." onClick="return false;">
                         Exporting Figure
-                        <span class="glyphicon glyphicon-refresh"></span>
+                        <span class="glyphicon glyphicon-refresh spinner"></span>
                     </button>
 
                     <a id="pdf_download" href="#" target="_blank"
--- a/omero_figure/static/figure/css/figure.css
+++ b/omero_figure/static/figure/css/figure.css
@@ -737,7 +737,7 @@
         100% { -webkit-transform: rotate(360deg); transform:rotate(360deg); }
     }
 
-    .navbar-right .glyphicon-refresh, .imgContainer .glyphicon-refresh {
+    .glyphicon-refresh.spinner {
         animation: spin 2s linear infinite;
         padding: 1px 2px 2px 2px;
     }
--- a/omero_figure/templates/figure/index.html
+++ b/omero_figure/templates/figure/index.html
@@ -833,7 +833,7 @@
                     <button id="pdf_inprogress" style="display:none" class="btn btn-primary btn-sm"
                             title="Figure generation in progress..." onClick="return false;">
                         Exporting Figure
-                        <span class="glyphicon glyphicon-refresh"></span>
+                        <span class="glyphicon glyphicon-refresh spinner"></span>
                     </button>
 
                     <a id="pdf_download" href="#" target="_blank"
--- a/src/js/models/roi_model.js
+++ b/src/js/models/roi_model.js
@@ -28,7 +28,7 @@ var ShapeModel = Backbone.Model.extend({
             }
         });
         // Convert other attributes
-        _.each(["Points", "MarkerEnd", "MarkerStart", "X", "Y", "RadiusX", "RadiusY", "X1", "Y1", "X2", "Y2", "Width", "Height", "TheZ", "TheT"], function(attr) {
+        _.each(["Points", "MarkerEnd", "MarkerStart", "X", "Y", "RadiusX", "RadiusY", "X1", "Y1", "X2", "Y2", "Width", "Height", "TheZ", "TheT", "TheC"], function(attr) {
             if (shape[attr] !== undefined) {
                 shape[lowerFirst(attr)] = shape[attr];
                 delete shape[attr];
--- a/src/js/views/roi_loader_view.js
+++ b/src/js/views/roi_loader_view.js
@@ -39,7 +39,7 @@ var RoiLoaderView = Backbone.View.extend({
             var shapeId = parseInt($tr.attr('data-shapeId'), 10);
             var shape = this.collection.getShape(shapeId);
             var shapeJson = shape.toJSON();
-            this.collection.trigger('shape_add', [shapeJson]);
+            this.collection.trigger('addShapesFromOmero', [shapeJson]);
         }
     },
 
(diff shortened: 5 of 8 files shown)


jburel:
General point: What about shapes that are linked to a specific channel?
Will the channel be turned on if the shapes are added and the channel is off?

pwalczysko:
Discussed with @will-moore and identified several issues:

Bugs:
1. There is no warning in case the single shape which you are attempting to add is outwith the viewport. Atm, every shape has a green enabled button, whether it is within the viewport or not, and this button is "clickable" even for shapes outwith the viewport, with no action and no message about the ROIs not being added
2. Variation on the Bug 1. above, the Add All button does not have any warning either, in case there are no Shapes on the particular z or t plane - the button is active and "clickable", but no message about the futility of the action comes forth.

RFE:
1. Explain somehow the fact that only 500 ROIs are loaded even when the "Add All" button is clicked, because of the loading by pages behaviour.

After consultation, both the bugs and the RFE could be helped a lot using a popup message after you have added a ROI saying "Added so-and-so many ROIs".

Also, it would be of value to ask on the original image.sc thread whether an "Add All" button which adds all ROIs from all z and t planes was meant, or, if the additon of the Shapes from the current z/t plane (as implemented now) is enough.

will-moore:
Improved messages for the "Add All" button, e.g:

Screenshot 2021-04-22 at 00 26 08

Screenshot 2021-04-22 at 00 20 38

Screenshot 2021-04-22 at 00 30 33

(also fixed the existing message for the "Add" button on each single Shape).

will-moore:
As noted in a response on the original image.sc topic, adding Shapes ONLY on the current Z/T plane is sufficient.

will-moore:
Simplest success message for `Add All`:

Screenshot 2021-04-22 at 15 49 34

As discussed, if `Add All` adds a Shape linked to a channel that is inactive, we turn on that channel(s).
Also let the user know if we're only adding shapes from 1 page of ROIs and if some shapes are not added because they lie outside the viewport:

Screenshot 2021-04-22 at 15 42 32

To test the channels workflow, you can add `theC` to a shape (first finding the ID of a shape in iviewer)
```$ omero obj update Polygon:439071 theC=1```
Then turn that channel off in figure before "Add All".
Adding just 1 shape with `theC` set will also activate the channel but no dialog shown (same logic as for Z/T).

pwalczysko:
Tested the Add and Add All buttons with shapes within and without the highlighted region, like the behaviour, the dialogs make sense.

The only tiny thing would be in the case where the "Add" fails to use
1. RFE
``Could not`` instead of ``Couldn't``

I will have a look at the channel behaviour on Monday.

pwalczysko:
Tested the channels behaviour on https://merge-ci.openmicroscopy.org/web/webclient/?show=image-140171 user-3.
All works as described, just the formulation on the dialog which reports about the ROIs addition could be improved.

``Shapes linked to an inactive channel (2) were added, so this channel has been turned ON.`` ->
``As shapes linked to an inactive channel (2) were added, this channel has been turned ON.``

pwalczysko:
In summary, all works as expected here, just two RFEs 
1. https://github.com/ome/omero-figure/pull/430#issuecomment-825785950
2. https://github.com/ome/omero-figure/pull/430#issuecomment-827475195

will-moore:
That previous commit fails the build with ```//omero-figure/.omeroci/app-deps: line 6: npm: command not found```.
Reading https://linuxize.com/post/how-to-install-node-js-on-centos-7/ adding `sudo yum install nodejs`...

pwalczysko:
Checked the messages, all good. The PR is green and the functionality is also as epected. 
Ready to merge fmpov.

jburel:
@will-moore could you put the 4 commits starting from e866aeeedf37a0538e629634eaf528b99581b89f in a separate PR? They not related to the feature worked on in this PR. This will be much easier to track down later
Thanks

will-moore:
@jburel Removed those commits from this branch and moved them to #435

jburel:
Just an idea, we have example of images with large number of ROIs. Was it tested with a very large number of rois/shapes?

pwalczysko:
> Just an idea, we have example of images with large number of ROIs. Was it tested with a very large number of rois/shapes?

2362 ROIs in total on https://merge-ci.openmicroscopy.org/web/webclient/?show=image-140171 (but not all the ROIs are on the same plane)

will-moore:
It's been tested on images with about 100 ROIs.
It only adds the Shapes from the current page of ROIs, so the max we'd expect to handle is 500 ROIs, although the number of Shapes that are added could be more or less than this, since we only pick Shapes that are on the current Z/T plane, and then only if they are within the viewport.

pwalczysko:
Did a new test with 
150086
ROIs. The image https://merge-ci.openmicroscopy.org/web/webclient/?show=image-105665 is a large one.
Lodded ROIs from 2 pages when zoomed in, and from one page (nr. 17) when fully zoomed out.
Both worked, and the ROIs appeared as expected.

pwalczysko:
@will-moore 

> It's been tested on images with about 100 ROIs.

see https://github.com/ome/omero-figure/pull/430#issuecomment-849764161 and https://github.com/ome/omero-figure/pull/430#issuecomment-849780713

pwalczysko:
The figure is https://merge-ci.openmicroscopy.org/web/figure/new/?image=105665

jburel:
which user?

will-moore:
user-3

jburel:
Things are slowly coming back
Screenshot 2021-05-28 at 16 23 54

jburel:
Screenshot 2021-05-28 at 17 13 09
After clicking "add all shapes"

will-moore:
I think those commits have made an improvement.
Hopefully it will be a bit faster and should be less likely to crash now.

jburel:
* Load the ROIS the first time. (it will also be good to have a spinner)
* Click add ``all shapes``: 200 shapes are added.
* Click ``refresh rois``, while this is happening, I click on ``Add all``
Screenshot 2021-06-01 at 13 07 58

PR #436, opened 2021-05-25
Update changelog for 4.5.0

will-moore:
cc @jburel


jburel:
I think it is good to mention https://github.com/ome/omero-figure/pull/435
This is not user facing but important from dev point of view

PR #464, opened 2022-05-20
cap reportlab

jburel:
Cap reportlab 
similar to what we did in our ci-infra
Check that the build is green


will-moore:
Thanks 👍

PR #468, opened 2022-06-08
Use 0.01 increment for channel sliders

### LINKED ISSUE #467: Fractional LUT range entries for float pixel values
For images with floating point pixel values, the iviewer allows one to enter fractional values for LUT min and max. Sometimes those ranges are small fractions. For example, MMP12 (Magenta) in this screenshot ranges from 0 to 0.6.

Screen Shot 2022-06-03 at 3 21 40 PM

The Omero figure interface does not properly check for floating point pixel values. Min and max accept only integer values (no decimal point). For small fractional values, the LUT display range in OMERO-figure is unable to cope.

Screen Shot 2022-06-03 at 3 21 20 PM

This image was taken with an Akoya PhenoImager. Their spectral unmixing produces float pixels.

will-moore:
Fixes #467.

 To test - see https://merge-ci.openmicroscopy.org/web/figure/file/95252 for example figure and images (user-3)

- Channel sliders should now have an increment of `0.01` allowing greater control, particularly with floating-point images with small ranges
- When you type values into the start/end input fields, they are no-longer forced to be Integers:

Screenshot 2022-06-10 at 10 59 33

### DIFF
--- a/src/js/views/channel_slider_view.js
+++ b/src/js/views/channel_slider_view.js
@@ -282,6 +282,7 @@ var ChannelSliderView = Backbone.View.extend({
                     range: true,
                     min: min,
                     max: max,
+                    step: 0.01,
                     values: [startAvg, endAvg],
                     slide: function(event, ui) {
                         $('.ch_start input', $div).val(ui.values[0]);
--- a/src/js/views/channel_slider_view.js
+++ b/src/js/views/channel_slider_view.js
@@ -1,4 +1,7 @@
 
+const SLIDER_INCR_CUTOFF = 100;
+// If the max value of a slider is below this, use smaller slider increments
+
 var ChannelSliderView = Backbone.View.extend({
 
     template: JST["src/templates/channel_slider_template.html"],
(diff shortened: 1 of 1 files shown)


dominikl:
It works. But a downside is that it also applies to int type images.

will-moore:
Since we don't store pixel-type info for images in figure, I've used the maximum pixel value of the range slider to determine if the slider increments should be small `0.01` or integers. The current threshold is 100. Below this we use `0.01` for slider increment.
I think this mostly behaves OK. It means you can have different behaviour for different channels of an Image, but if one channel is quite dark, that's probably OK.

jburel:
@will-moore is it possible to make the textbox a bit bigger to avoid the truncation?

will-moore:
@jburel When are you seeing this truncation? I can't reproduce, since we only allow non-integer values in the text boxes if the max intensity is less than 100, and then it's formatted to 2 decimal places, so a max of 4 characters, which looks OK:

Screenshot 2022-07-01 at 13 44 18

will-moore:
In discussion with @jburel, the truncation he was seeing was in the first screenshot above, *before* https://github.com/ome/omero-figure/pull/468/commits/5f0bf0f798687e69108dabb4b66d95c857b14f06 added the formatting of those text inputs. So I think this is good to merge?

PR #471, opened 2022-07-06
Use thumbnail if scaled region too large

### LINKED ISSUE #449: Bug: render scaled region larger than max_size
The longest size will exceed the max_size given the logic at
See https://github.com/ome/omero-figure/blob/b31ace415ac170b970b88f26311f83f1fdb42436/omero_figure/views.py#L183

This can cause a much larger region than needed to be rendered (especially if there are a small number of resolution levels) causing memory issues. See
https://forum.image.sc/t/omero-figure-memory-usage-when-loading-slidescans/58322

will-moore:
Fixes #449 

To test:
 - Need big images with not many resolution levels, so that the smallest level has more than e.g. 5k * 5k pixels (configurable).
 - Zooming out of the figure panel, so that a large region of the image is rendered will trigger the thumbnail to be used. Zooming in will allow a small area to be rendered.

### DIFF
--- a/omero_figure/views.py
+++ b/omero_figure/views.py
@@ -43,6 +43,7 @@ from omeroweb.webclient.decorators import login_required
 from .omeroutils import get_timestamps
 
 from . import utils
+from . import figure_settings
 import logging
 
 try:
@@ -173,6 +174,7 @@ def render_scaled_region(request, iid, z, t, conn=None, **kwargs):
     if scale_levels is None:
         # Not a big image - can load at full size
         level = None
+        scale = 1
     else:
         # Pick zoom such that returned image is below MAX size
         max_level = len(scale_levels.keys()) - 1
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -106,6 +106,25 @@ Now install the script's dependencies:
...
 
+
+OMERO.figure settings
+---------------------
+
+max_rendered_region
+^^^^^^^^^^^^^^^^^^^
+
+When displaying a big (tiled) image, OMERO.figure will load data from an appropriate
+level of the tiled pyramid, accoring to the region to be displayed. However, if the
```
jburel, line 117:
```suggestion
level of the tiled pyramid, according to the region to be displayed. However, if the
```

omero_figure/views.py
```diff
@@ -188,33 +190,51 @@ def render_scaled_region(request, iid, z, t, conn=None, **kwargs):
...
+            paste_y = int(-new_y)
+            new_height = new_height + new_y
+            new_y = 0
 
     # Render the region...
-    jpeg_data = image.renderJpegRegion(z, t, x, y, width, height, level=level,
-                                       compression=compress_quality)
+    if (new_width * new_height) > figure_settings.MAX_RENDERED_REGION:
+        # If region is too big, use thumbnail instead...
+        thumb_data = image.getThumbnail(size=96)
```
jburel, line 222: shouldn't the default value from server be used?

omero_figure/views.py
```diff
@@ -188,33 +190,51 @@ def render_scaled_region(request, iid, z, t, conn=None, **kwargs):
...
+        thumb_img = Image.open(BytesIO(thumb_data))
+        # resize the thumb to the scaled full image size
+        thumb_img = thumb_img.resize((new_size_x, new_size_y))
+        # crop to resized coordinates if necessary
+        if (new_x > 0 or new_y > 0 or new_width < new_size_x
+                or new_height < new_size_y):
+            thumb_img = thumb_img.crop((new_x, new_y, new_x + new_width,
+                                        new_y + new_height))
+        rv = BytesIO()
```
jburel, line 231: shouldn't it be closed?

will-moore:
Current blocker on testing is the import of 2.5 GB file from https://forum.image.sc/t/omero-figure-memory-usage-when-loading-slidescans/58322/11 into merge-ci. Transfer takes over 2 hours (importing from local machine via CLI from within UoD) and ends up failing due to connection loss.

pwalczysko:
@will-moore https://merge-ci.openmicroscopy.org/web/webclient/?show=dataset-58919 user-3

omero_figure/views.py
will-moore: fixed in [3fbd7c2](https://github.com/ome/omero-figure/pull/471/commits/3fbd7c2760f21706bdf61bbf62cf158c6aa52aad)

omero_figure/views.py
will-moore: fixed in fd4fcca69

pwalczysko:
First testing shows a problem:

1. Open figure https://merge-ci.openmicroscopy.org/web/figure/new/?image=243715 (user-3).
2. Observe that the panel is zoomed out fully, do not change this.
3. Change some rendering settings in the Preview pane, either switch a channel off or slide a Min/Max slider.
4. Observe that there is a placeholder for couple of seconds (two arrows chasing each other) on the panel.
5. After the placeholder is gone, the panel is white -> not expected, the expected behaviour is that the rnd settings do not change, as they should be the ones taken from OMERO.
6. The inspector in FF is showing that the thumbnail is returned as expected, but the experience as described in Ad 5. is not good, suggesting that the returned thumbnail is not being displayed.
7. Do not zoom in, or change rnd settings now, but drag the panel to make it bigger.
8. Observe that the dragging in Ad. 7 fixed the problem in Ad 5. -> the panel is no more white, but the thumbnail is displayed, and the rnd settings are the same as in OMERO.
9. Zoom in and change rnd settings -> observe that the experience is as expected, as long as you zoomed in sufficiently (i.e. zoomed in enough to pass the region where the thumbnail is displayed, and instead, the first subresolution is displayed)

will-moore:
@pwalczysko it seems those issues are limited to Firefox which doesn't handle a big images very well. That last commit should improve things.

pwalczysko:
This works fine now on FF.
Lgtm.

will-moore:
For testing the config, I have added to merge-ci:
```
$ omero config set omero.web.figure.max_rendered_region 1000000
```
This means that at the smallest pyramid level, if the rendered region is above 1000 x 1000 pixels, then use the thumbnail. The default (tested above) is 5000 x 5000. So with this config, you expect to have to zoom in a bit further before the thumbnail is replaced by the lowest resolution of the pyramid...

pwalczysko:
In FF, on MacOS, I have a 504 on the Figure in the description.
See the url below which does not return:

https://merge-ci.openmicroscopy.org/web/figure/render_scaled_region/243715/0/0/?c=1|0:183$FF0000,2|0:255$00FF00,3|0:255$0000FF&maps=[{%22reverse%22:{%22enabled%22:false}},{%22reverse%22:{%22enabled%22:false}},{%22reverse%22:{%22enabled%22:false}}]&region=92730,38650,1034,1034&m=c

Workflow:
Open the Figure, switch channels on and off, manipulate the sliders. Then, zoom in a bit, so that you do not see the thumbnail, but the higher sub-resolution (the image in the Preview is well-resolved, not just the pixelated overzoomed thumb). Then, drag the lower bar on the panel down, to make the viewport bigger (on the figure panel). This causes endless spinner.

will-moore:
This is conflicting with #477 so closing for now...

Issue #318, opened 2018-12-19
Labels for x/y/z


joshmoore:
When zooming and when showing multiple Z slices (see recent 1000+ z-slide volumes) being able to label with the xyz coordinates would be useful.

joshmoore:
@will-moore : this was something else I ran into while trying to create https://github.com/ome/omero-figure/issues/319#issuecomment-454730333

joshmoore:
![screen shot 2019-01-16 at 11 36 19](https://user-images.githubusercontent.com/88113/51243514-111c7a80-1983-11e9-83ca-f244e22077a1.png)

will-moore:
If I were to choose ```x``` or ```y``` then the label would show the current coordinate of the top-left corner of the viewport?
This is dynamic, since it can change when you change the shape of the panel. E.g if you resize a panel from a horizontal rectangle to a vertical rectangle then the top-left corner is at a different point (even if you don't pan). Would you want a single label for coordinates e.g. ```x: 100, y: 250```? The current behaviour for T-index is to show a single number (e.g. ```10``` not ```T: 10```) so if we did the same for Z then you could end up with labels (one per row) like:

```
10
5
```
instead of 
```
T: 10
Z: 5
```
So maybe we need to re-think the existing T-index behaviour.

PR #472, opened 2022-07-11
Dynamic labels with regex

Tom-TBT:
PLEASE READ UNTIL THE END BEFORE MERGING (Figure_To_Pdf.py not tested and missing part about JSON version compatibility)

Dear OME team,
this pull request concerns labels, and in particular how special values like time, X/Y/Z or image name are handled.

Before, only "time labels" that were specified like [time-hrs:mins:secs] would be dynamically handled . But that worked only if the text of the label was exactly starting like "[time" and ending with an understood format like "-hrs:mins:secs]".

I changed the behavior so that a regex looks for a value between square brackets [], split the content separated by a "-" character, and replaces the text in the label according to what it found. With all the commits from this PR, the code recognizes:

* [time-...]
  - index
  - milliseconds
  - ... (all that existed before)
 * [name-...]
   - image (that was given before as image-name)
   - dataset (that was given before as dataset-name)
 * [roi-...] (same as the viewport values)
   - x
   - y
   - width
   - height
   - rotation (bonus)
* [depth-...]
  - index
  - unit (display the depth in units, like 17.6 μm)

grafik

 
In my opinion, it simplifies the code, were the "[time" special label was handled in many places and thus was harder to extend. In addition, it permits three new things:
* Text and dynamic labels can be mixed
* Several dynamic labels can be used in a single label
* Markdown can be used on those dynamic labels (and on the text surrounding them)

--------------------------------------------------------------

From there, if you want to merge my updates, we need first to:
- Test Figure_To_Pdf.py (I haven't found a way yet to test it)
- Handle the bumped version of the JSON (src/js/models/figure_model.js). I had to add "data.pixel_size.z" for the depth, but I don't know where to get it from in "version_transform"
- Review the naming of the dynamic labels -> I'm not a great name picker for variables, and I kept what was existing for the parsing (square brackets and values separated by "-"). Maybe you want to change it to something more intuitive now, because it will be annoying to change it in the future

--------------------------------------------------------------

Finally, I found this issue (https://github.com/ome/omero-figure/issues/318) that could be solved by this PR 

Thanks for your thoughts/critics/feedback. I'm also not so experienced in collaborating on Github so I apologize in advance if things are a bit messy here.

Cheers!
Tom

### DIFF
--- a/omero_figure/views.py
+++ b/omero_figure/views.py
@@ -131,9 +131,10 @@ def img_data_json(request, image_id, conn=None, **kwargs):
         rv['pixel_size']['symbolY'] = py.getSymbol()
         rv['pixel_size']['unitY'] = str(py.getUnit())
         pz = image.getPrimaryPixels().getPhysicalSizeZ()
-        rv['pixel_size']['valueZ'] = pz.getValue()
-        rv['pixel_size']['symbolZ'] = pz.getSymbol()
-        rv['pixel_size']['unitZ'] = str(pz.getUnit())
+        if pz is not None:
+            rv['pixel_size']['valueZ'] = pz.getValue()
+            rv['pixel_size']['symbolZ'] = pz.getSymbol()
+            rv['pixel_size']['unitZ'] = str(pz.getUnit())
     size_t = image.getSizeT()
     time_list = []
     if size_t > 1:
--- a/omero_figure/views.py
+++ b/omero_figure/views.py
@@ -130,6 +130,10 @@ def img_data_json(request, image_id, conn=None, **kwargs):
         rv['pixel_size']['valueY'] = py.getValue()
         rv['pixel_size']['symbolY'] = py.getSymbol()
         rv['pixel_size']['unitY'] = str(py.getUnit())
+        pz = image.getPrimaryPixels().getPhysicalSizeZ()
+        rv['pixel_size']['valueZ'] = pz.getValue()
+        rv['pixel_size']['symbolZ'] = pz.getSymbol()
+        rv['pixel_size']['unitZ'] = str(pz.getUnit())
     size_t = image.getSizeT()
     time_list = []
     if size_t > 1:
--- a/omero_figure/views.py
+++ b/omero_figure/views.py
@@ -157,6 +157,22 @@ def timestamps(request, conn=None, **kwargs):
 
 
 @login_required()
+def z_scale(request, conn=None, **kwargs):
+
+    iids = request.GET.getlist('image')
+    data = {}
+    for iid in iids:
+        image = conn.getObject('Image', iid)
+        if image is not None:
+            pz = image.getPrimaryPixels().getPhysicalSizeZ()
+            if pz is not None:
+                data[image.id] = {'valueZ': pz.getValue(),
+                                  'symbolZ': pz.getSymbol(),
+                                  'unitZ': str(pz.getUnit())}
+    return JsonResponse(data)
+
+
+@login_required()
 def render_scaled_region(request, iid, z, t, conn=None, **kwargs):
 
     region = request.GET.get('region')
--- a/src/js/models/panel_model.js
+++ b/src/js/models/panel_model.js
@@ -354,11 +354,12 @@
                     start = this.get('z_start');
                     end = this.get('z_end');
                     if (format === "pixel") {
-                        text = "" + (start+1) + "-" + (end+1);
+                        text = "" + (start+1) + " - " + (end+1);
                     } else if (format === "unit"){
                         start = (start * this.get('pixel_size_z')).toPrecision(3)
                         end = (end * this.get('pixel_size_z')).toPrecision(3)
-                        text = ""+ start + "-" + end +" "+ this.get('pixel_size_z_symbol')
+                        text = ""+ start +" "+ this.get('pixel_size_z_symbol')
+                               + " - " + end +" "+ this.get('pixel_size_z_symbol')
                     }
                 }
                 else {
--- a/src/js/models/panel_model.js
+++ b/src/js/models/panel_model.js
@@ -350,12 +350,26 @@
 
             var text = "";
             if (property === "z") {
-                var theZ = this.get('theZ');
-                if (format === "pixel") {
-                    text = "" + (theZ + 1);
-                } else if (format === "unit"){
-                    text = ""+ (theZ * this.get('pixel_size_z')).toPrecision(3) +" "+ this.get('pixel_size_z_symbol')
+                if (this.get('z_projection')) {
+                    start = this.get('z_start');
+                    end = this.get('z_end');
+                    if (format === "pixel") {
+                        text = "" + (start+1) + "-" + (end+1);
+                    } else if (format === "unit"){
+                        start = (start * this.get('pixel_size_z')).toPrecision(3)
+                        end = (end * this.get('pixel_size_z')).toPrecision(3)
+                        text = ""+ start + "-" + end +" "+ this.get('pixel_size_z_symbol')
+                    }
                 }
+                else {
+                    var theZ = this.get('theZ');
+                    if (format === "pixel") {
+                        text = "" + (theZ + 1);
+                    } else if (format === "unit"){
+                        text = ""+ (theZ * this.get('pixel_size_z')).toPrecision(3) +" "+ this.get('pixel_size_z_symbol')
+                    }
+                }
+
                 return text
             }
             viewport = this.getViewportAsRect();
--- a/src/js/models/panel_model.js
+++ b/src/js/models/panel_model.js
@@ -94,11 +94,14 @@
                 'orig_width': data.orig_width,
                 'orig_height': data.orig_height,
                 'datasetName': data.datasetName,
+                'datasetId': data.datasetId,
                 'pixel_size_x': data.pixel_size_x,
                 'pixel_size_y': data.pixel_size_y,
-				'pixel_size_z': data.pixel_size_z,
+                'pixel_size_z': data.pixel_size_z,
                 'pixel_size_x_symbol': data.pixel_size_x_symbol,
+                'pixel_size_z_symbol': data.pixel_size_z_symbol,
                 'pixel_size_x_unit': data.pixel_size_x_unit,
+                'pixel_size_z_unit': data.pixel_size_z_unit,
                 'deltaT': data.deltaT,
             };
 
@@ -291,27 +294,23 @@
                 isNegative = (deltaT < 0),
                 text = "", h, m, s;
             deltaT = Math.abs(deltaT);
+            h = parseInt(deltaT / 3600);
+            m = parseInt(deltaT / 60);
+            s = pad(Math.round(deltaT % 60));
             if (format === "index") {
                 isNegative = false;
                 text = "" + (theT + 1);
-            } else if (format === "milliseconds") {
+            } else if (['milliseconds', 'ms'].includes(format)) {
                 text = Math.round(deltaT*1000) + " ms";
-            } else if (format === "secs") {
+            } else if (['seconds', 'secs', 's'].includes(format)) {
                 text = Math.round(deltaT) + " s";
-            } else if (format === "mins:secs") {
-                m = parseInt(deltaT / 60);
-                s = pad(Math.round(deltaT % 60));
-                text = m + ":" + s;
-            } else if (format === "mins") {
+            } else if (['minutes', 'mins', 'm'].includes(format)) {
                 text = Math.round(deltaT / 60) + " mins";
-            } else if (format === "hrs:mins") {
-                h = parseInt(deltaT / 3600);
-                m = pad(Math.round((deltaT % 3600) / 60));
+            } else if (["mins:secs", "m:s"].includes(format)) {
+                text = m + ":" + s;
+            } else if (["hrs:mins", "h:m"].includes(format)) {
                 text = h + ":" + m;
-            } else if (format === "hrs:mins:secs") {
-                h = parseInt(deltaT / 3600);
-                m = pad(parseInt((deltaT % 3600) / 60));
-                s = pad(Math.round(deltaT % 60));
+            } else if (["hrs:mins:secs", "h:m:s"].includes(format)) {
                 text = h + ":" + m + ":" + s;
             } else { // Format unknown
                 return ""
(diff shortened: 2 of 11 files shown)


jburel:
```
flake8.main.application   MainProcess   1266 INFO     Found a total of 63 violations and reported 1
./omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py:1113:14: E261 at least two spaces before inline comment

```

jburel:
@Tom-TBT Did you try to expand https://github.com/ome/omero-figure/blob/master/test/integration/test_figure_scripts.py?
The tests are run by the GitHub action against an ``omero-server``

src/templates/labels_form_inner_template.html
```diff
@@ -42,6 +42,15 @@
...
                         <a href="#" data-label="[time-hrs:mins]">Time (hrs:mins)</a>
+                    </li>
+					<li class="add_depth_label"  title="Add a label that shows the current Z-index">
+                        <a href="#" data-label="[depth-index]">Depth (Z-index)</a>
+                    </li>
+                    <li class="add_depth_label"  title="Add a label that shows the current depth">
+                        <a href="#" data-label="[depth-unit]">Depth (unit)</a>
+                    </li>
+					<li class="add_roi_label"  title="Add a label that shows the viewport X and Y">
+                        <a href="#" data-label="X: [roi-x] Y: [roi-y]">X and Y (pixel)</a>
```
will-moore, line 53: I think using `roi` here could be confusing since there's no ROI on the image (even more confusing if there was an ROI). Maybe use e.g. `[viewport-x]` or if that's too long then `[view-x]`?

Tom-TBT: Viewport/view sounds better than roi. We could also make aliases, to have one explicit name and one faster to type:
-> viewport : view : v
->time : t
->depth : d
->channels : c
-> name : n

src/js/models/panel_model.js
```diff
@@ -301,25 +313,56 @@
...
+            if (format === "image") {
+                var pathnames = this.get('name').split('/');
+                text = pathnames[pathnames.length-1];
+            } else if (format === "dataset"){
+                text = this.get('datasetName') ? this.get('datasetName') : "No/Many Datasets";
+            }
+            return text;
+        },
+		
+		get_roi_label_text: function(format) {
```
will-moore, line 336: The indentation looks wrong here - I think you have Tabs instead of Spaces. Same in a bunch of other places. It looks OK in my editor but still best to keep it consistent - thx

will-moore:
Hi @Tom-TBT and thanks for the PR. I like the idea of using a regex to dynamically create labels, allowing users to add their own text to the dynamic tokens.

- You should be able to test the `Figure_To_Pdf.py` quite easily by uploading the script under `omero/figure_scripts/`. Then, omero-figure will find it there and use it for export. If you already have the original script there, simply replace it with your edited script from this PR
- I tried opening a previous figure, with time-stamp labels (that are saved in the format `{"time": "mins", ...}` instead of the as in this PR `{"text": "[time-mins]", ...}`. This fails since the regex is looking for a "text" attribute, which is missing. You'll need to update the JSON at `if (v < 6) {...` when you open the older format.
- This would also be the place to do other updates. E.g. if you go for `name-dataset` instead of `dataset-name` that would need to be handled here, although if you go for `object-attribute` (as in `viewport-x`) then `dataset-name` or `dataset-id` would be most consistent (and wouldn't be a breaking change).
- However, if we went for a breaking change then we could also use a different syntax e.g. with a `.` like `dataset.name`, `viewport.x` or `time.mins:secs`. 
- The current handling of `[channels]` (creates separate channel labels) or e.g. `My [channels]` creates single dynamic channels label, isn't intuitive or discoverable, although I can see the use-case. Maybe add a new drop-down option `Channels (single label)` with a value `[channels]` (which gives the new behaviour without needing extra text) and the existing `Channels` option has some other value e.g. `[channel labels]` that creates the separate labels?
- For pixel-size-Z support, you can add Z handling at https://github.com/ome/omero-figure/blob/ee3f99d9809f085696e72524bf296343345d5fb3/omero_figure/views.py#L132 (as it is above for X and Y) to give you all the Z info in the image JSON (instead of having to assume that Z units are the same as X units). Then you can include this in the figure JSON as you've already done for pixel size z. However, in order to allow a user to choose `depth-unit` for pre-existing figures, you'll need to dynamically update the JSON when a user opens an old figure (as is done for time-stamps for version 5). If this looks too tricky to handle just now, we could leave it to a follow-up PR, or at least leave it until everything else is working.
- It could be nice to handle max-intensity projection for "depth" e.g. "1.0 µm - 2.5 µm".
- Actually, I'm not sure about the term "depth", since this isn't used anywhere else in OMERO. Maybe you could also use "viewport" or "view" here, since it's the same concept as X and Y (do you want to support 'unit' lengths for X and Y, as you do for Z? E.g. `viewport-x-pixel, viewport-x-unit, viewport-z-unit, viewport-z-pixel` - you could even drop the "viewport" from all of these options!?

OK, I'll stop for now. Apologies for the evolving ideas during this review - lots to think about!
Cheers,
 Will

Tom-TBT:
Hi @will-moore,
I haven't added the code for compatibility with the previous version yet, but here is my progress:
* Removed tabs (all of them hopefully)
* I changed the parsing from `[object-attribute]` to `[object.attribute]`. 
* Attributes have default values when writing `[object]`:
  *  time -> index
  * image/dataset -> name
  * x/y/z/width/height -> pixel
* Labels for : `[image.id]`, `[image.name]`, `[dataset.id]`, `[dataset.name]`
* Harmonization of `x`, `y`, `z`, `width`, `height` to be displayed either as `unit` or `pixel`
* Rotation displayed with the symbol, like `78°`
* Aliases object: 
  * `time` / `t`
  * `rotation` / `rot`
  * `width` / `w`
  * `height` / `h`
* Aliases attribute: 
  * `milliseconds` / `ms`
  * `seconds` / `secs` / `s`
  * `minutes` / `mins` / `m`
  * `mins:secs` / `m:s`
  * `hrs:mins` / `h:m`
  * `hrs:mins:secs` / `h:m:s`
  * `pixel` / `px`
* Two drop-down options for channels: one old behavior [channels labels] and one new [channels]
* More drop-down option. If too much, we need to decide which ones to keep, which ones to add
* Z-projection is now handled. The label is the same, the switch happens on the display only. 
* I added the Z pixel sizes and symbols to the model


Regarding testing `Figure_To_Pdf.py`, my issue was that after replacing the script, I had an error saying that the file hash changed and refused to execute. But it works after reloading OMERO.web and opening a new OMERO.figure tab.
For the workflow, I use the cli to upload my script on the server with a console opened in my container:
```
. /opt/omero/web/venv3/bin/activate
cd omero_figure/scripts/omero/figure_scripts
omero script upload Figure_To_Pdf.py --official
```
(I haven't found yet the command replacing directly `omero/figure_scripts/Figure_To_Pdf.py`)

Let me know if you have more ideas or updates in mind!
Tom

will-moore:
Wow, thanks.
I'm away next week, so I might not test again till after that.

Try to replace script by getting the ID (only have to do this once) then using `replace`:

```
omero script list
omero script replace ID omero/figure_scripts/Figure_To_Pdf.py
```

I think we need to be conservative with the new features that you're adding and only add what you really need, because once you've released it (and exposed it in the UI) and people are using it (or *might* be using it) then we have to support it test it, document it etc going forwards.
If some of this functionality isn't exposed in the UI (no drop-down option) it could still be tested or used as a "hidden feature" but we wouldn't be considered a breaking change for users if we decided not to support it in future.

omero_figure/views.py
```diff
@@ -130,6 +130,10 @@ def img_data_json(request, image_id, conn=None, **kwargs):
         rv['pixel_size']['valueY'] = py.getValue()
         rv['pixel_size']['symbolY'] = py.getSymbol()
         rv['pixel_size']['unitY'] = str(py.getUnit())
+        pz = image.getPrimaryPixels().getPhysicalSizeZ()
+        rv['pixel_size']['valueZ'] = pz.getValue()
```
will-moore, line 134: You'll need to check `if pz is not None:` before using it, since for some 2D images they may have pixel sizes for X and Y, but not for Z.

src/js/models/panel_model.js
```diff
@@ -279,47 +295,99 @@
...
+
+            var text = "";
+            if (property === "z") {
+                if (this.get('z_projection')) {
+                    start = this.get('z_start');
+                    end = this.get('z_end');
+                    if (format === "pixel") {
+                        text = "" + (start+1) + " - " + (end+1);
+                    } else if (format === "unit"){
+                        start = (start * this.get('pixel_size_z')).toPrecision(3)
```
will-moore, line 359: Need to handle `pixel_size_z` being undefined here too.

src/js/models/panel_model.js
```diff
@@ -279,47 +295,107 @@
...
-                m = pad(Math.round((deltaT % 3600) / 60));
+            } else if (["mins:secs", "m:s"].includes(format)) {
+                text = m + ":" + s;
+            } else if (["hrs:mins", "h:m"].includes(format)) {
                 text = h + ":" + m;
-            } else if (format === "hrs:mins:secs") {
-                h = parseInt(deltaT / 3600);
-                m = pad(parseInt((deltaT % 3600) / 60));
-                s = pad(Math.round(deltaT % 60));
+            } else if (["hrs:mins:secs", "h:m:s"].includes(format)) {
```
will-moore, line 314: You've lost the `pad()` on the mins here and on the mins for `hrs:mins` above. E.g. `0:00:00` now reads as `0:0:00`. NB: for `mins:secs` you don't want to `pad(m)`.

will-moore:
I noticed that the PDF export doesn't format the lengths to the same number of decimal places at in the browser (screenshot showing PDF overlaying the browser also shows the timestamp formatting issue mentioned earlier):

Screenshot 2022-08-01 at 12 28 05

will-moore:
It would be good to document the new behaviour (E.g. that you can use `My time: [time.mins]` etc) so that other users get the full benefit of these nice new features.
A good place to add this info in in the dialog that currently just describes the `markdown` behaviour (not sure if many people notice that this is a button anyway).

With the diff below, this could become a general labels `Info` button (or some other text if you prefer) that shows the dialog. Then you just need to update the dialog itself (title "Label Formatting" instead of "Markdown Formatting") to include a note about the dynamic labels options and an example or two.

Screenshot 2022-08-01 at 12 56 06

```
$ git diff
index eec7f13..8bd5b9e 100644
@@ -167,16 +167,6 @@
         padding: 3px 15px;
     }
 
-    .markdown-info {
-        padding: 3px 12px;
-        color: #aaa;
-        text-align: right;
-        padding-left: 40px;
-        background: url(../images/markdown_light32x20.png) 0% center no-repeat;
-    }
-    .markdown-info:hover {
-        background: url(../images/markdown_dark32x20.png) 0% center no-repeat;
-    }
     .legend-container .markdown-info {
         display: none;
     }
@@ -184,12 +174,6 @@
         display: block;
     }
 
-    #labelsTab .markdown-info {
-        height: 18px;
-        margin: 9px 9px 0;
-        float: left"
-    }
-
     /* hide appropriate collapse/hide button */
     .legend-collapsed .collapse-legend{
         display: none;
diff --git a/omero_figure/templates/figure/index.html b/omero_figure/templates/figure/index.html
index 51c2ee3..9f45a3d 100644
--- a/omero_figure/templates/figure/index.html
+++ b/omero_figure/templates/figure/index.html
@@ -1020,7 +1020,9 @@
                     <h5 style="float: left">Add Labels</h5>
 
                     <button type="button" class="btn btn-link markdown-info"
-                        title="Use Markdown formatting. Click for more info..."></button>
+                        title="Show label formatting options...">
+                        Info
+                    </button>
                     <div class="clearfix"></div>
                     <form class="new-label-form form-inline" role="form">
                     </form>
```

src/js/models/panel_model.js
Tom-TBT: Sorry, I thought I could handle it all at once, but I missed that pad part. I reverted my changes

src/templates/labels_form_inner_template.html
```diff
@@ -11,10 +11,15 @@
                 <ul class="dropdown-menu pull-right">
                     <li role="presentation" class="dropdown-header">Create Label(s) From:</li>
                     <li role="presentation" class="divider"></li>
-                    <li><a href="#" data-label="[image-name]">Image Name</a></li>
-                    <li><a href="#" data-label="[dataset-name]">Dataset Name</a></li>
+                    <li><a href="#" data-label="[image.name]">Image Name</a></li>
+                    <li><a href="#" data-label="[image.id]">Image ID</a></li>
+                    <li><a href="#" data-label="[dataset.name]">Dataset Name</a></li>
+                    <li><a href="#" data-label="[dataset.id]">Dataset ID</a></li>
```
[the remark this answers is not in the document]
Tom-TBT: I added quite a lot to the dropdown list and I agree that it's too long. I don't really need `id` so I removed it from the drop-down list. I can remove more like `Rotation (degree)`, one of `Z (pixel)`/`Z (unit)` and `X and Y (pixel)`/`X and Y (unit)`

They should be discoverable now anyway with the updated Info/Tip menu

But your previous comment remains. Do we decide to expose it to the users thus agreeing to support it in the future? Do you think the feature is mature enough?

Tom-TBT:
Hey @will-moore, thanks a lot for spotting those issues.
List of updates in last commits:
* Time padding is back (sorry for that)
* The format for float values now uses `.toFixed(2)`, which has the same behavior as the string formatting in python `f"{value:.2f}"`
* Fixed view units difference between JS and python (e.g. `X: 4.72` was `X: 4.70` due to a too early `parseInt` in JS) 
* Update of label upon resizing of the image (matters for x, y, width and height)
* I also spotted and fixed a bug I introduced due to my aliases, not displaying [time.secs]
* Modified the label doc window according to your suggestion (though I preferred `Tips` over `Info`, what do you think about it?)
* Added examples and list of accepted values for the labels in the doc. Hopefully it's not too much doc to the curious user.
* Removed image.id and dataset.id to the drop-down list. I can remove more

src/templates/labels_form_inner_template.html
```diff
@@ -11,10 +11,13 @@
                 <ul class="dropdown-menu pull-right">
                     <li role="presentation" class="dropdown-header">Create Label(s) From:</li>
                     <li role="presentation" class="divider"></li>
-                    <li><a href="#" data-label="[image-name]">Image Name</a></li>
-                    <li><a href="#" data-label="[dataset-name]">Dataset Name</a></li>
+                    <li><a href="#" data-label="[image.name]">Image Name</a></li>
+                    <li><a href="#" data-label="[dataset.name]">Dataset Name</a></li>
                     <li title="Create labels from active Channels on the selected panel">
-                        <a href="#" data-label="[channels]">Channels</a>
+                        <a href="#" data-label="[channels labels]">Channels (separate label)</a>
```
will-moore, line 17: Minor: Should be text: `Channels (separate labels)` (plural labels).

will-moore:
Thanks for those updates.
I'm happy to go with "Tips".
The dialog looks good, but I actually missed the lower section as I didn't realise the dialog was scrollable when I opened it (common problem of scrollbars being hidden by default).

Screenshot 2022-08-02 at 11 57 09

 It's possible to force the scrollbars to be visible via css, but an easier solution is to boost the height a bit so that it shows part of the next line. e.g. 
```
<div class="modal-body" style="max-height:500px">
```
Part of the reason I didn't realise that text was hidden was because what's shown is almost enough info. The syntaxes section below, that lists all the options is probably not needed, since we show the important options in the drop-down list.
Is the support of various alias terms for each property an important feature? Again, it's more functionality to support, test and document, so if users only know a single way of specifying e.g. `time.mins` then it keeps things simpler.

For the other options like Rotation, that's up to you what you feel is needed/useful, but I would only add what you need now as you can always add more later, but removing stuff is harder.

I would just say something like: "Certain properties can be inserted in the label (within square brackets) and are dynamically updated upon changes. See the drop-down options for supported properties". Then just list 3 or 4 examples in that table to illustrate the syntax and remove the 2nd table, which is harder to understand than the examples section. This is assuming that the combination of drop-down list and the first table of examples can cover all the options we want to support?

Tom-TBT:
Ok to remove the detailed doc, and I'll make the features findable either in the dropdown or in the example.

About the alias, I could avoid it but I'm more a fan of the "short" version rather than the verbose one. It wasn't a problem when the label was chosen from the list (like `[time.hrs:mins:secs]`, but now that it can be combined with text, I expect it typed manually more often. The second point is that the label text-field is also small, so the short version could spare some scrolling.
Because replacing `hrs:min:secs` by `h:m:s` would be a breaking change from v5, I would have to implement it in the conversion to v6 (if I go for short). And some might prefer the verbose version for clarity.... That's why I thought alias would be a good compromise.

About the rotation, I thought of adding it because it wasn't much different from x/y/height/width (the difference is the ° symbol) and I wanted to add different "relevant" features. I don't have the use for it but someone else could.

Tom-TBT:
Hey Will,
here are the changes we talked about. 

I chose the odd value of `518px` for the dialog height to have a line cropped in half. Do you see the same thing on your side?
image

will-moore:
Opened a fix for the "Legend" button issue noticed today: https://github.com/ome/omero-figure/pull/474

will-moore (approved):
Looks great, thanks.

pwalczysko:
Thank you @Tom-TBT , very nice addition.

All works as expected. 
The only thing which I would invite one more thought is the question "How does a user keep an overview, how does he/she know what feature is dynamic and which is not ?"
Screenshot 2022-08-09 at 11 16 04
Could we maybe explain that a bit in the paragraph above ?
Also, I would remove the last sentence from the one in the screenshot, and add

``See the drop-down options for all supported properties. The table below lists some additional properties not included in the drop-down options.``

pwalczysko (changes requested):
Please comment on https://github.com/ome/omero-figure/pull/472#issuecomment-1209212060

pwalczysko:
> leave it like this for now

@Tom-TBT would be what is needed for now imho, but thanks for explaining the details. 

I think what would help (and that was the scope of my comment) is to explain the general situation to a naive user in most generic terms (of course only as much as possible). The dynamic vs non-dynamic is definitely too complex to even start on this, so I suggest we leave it (I understood, but the naive user would be confused).

We discussed with @will-moore an addition to the text in the "Tips". @will-moore is going to post exact text soon for you to okay it.

will-moore:
We discussed a different way of describing the dynamic labels that more closely matches the expected workflow...

"The drop-down menu in the 'Add Labels' form allows you to create labels based on Image metadata. Labels created with square brackets indicate dynamic properties. The properties within the square brackets will be dynamically updated upon changes. Additional text can be added outside the square brackets and will not be affected when the labels are displayed.
The following table shows some examples, including the use of dynamic properties and markdown together".

Hope that sounds OK to you?

pwalczysko (approved):
Thank you again @Tom-TBT !

will-moore:
Hi @Tom-TBT - There'll be a link to this PR in the release changelog - for those wanting more info on what's new...
I wonder if you could update the description of the PR since a few things changed (as @pwalczysko found when he started testing, e.g. "depth", "name-..." etc.) Thanks.

PR #474, opened 2022-08-08
Only show legend-container above right panel when editing

will-moore:
This fixes a minor issue where the right-panel (or part of it like a drop-down list) falls behind the legend button (see screenshot). Only happens when the screen is small and/or the drop-down list is longer (see #472):

Screenshot 2022-08-08 at 14 10 46

With the fix, the legend controls and the legend itself only shows above the right panel when you are editing the legend.

### DIFF
--- a/omero_figure/static/figure/css/figure.css
+++ b/omero_figure/static/figure/css/figure.css
@@ -139,7 +139,7 @@
 
     /* -- Legend -- */
     .legend-container {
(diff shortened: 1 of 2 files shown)


pwalczysko:
Tested on merge-ci, compared with latest-ci where I was able to repeat the bug.

LGTM

pwalczysko (approved):

PR #480, opened 2022-09-13
Multi label text fix

Tom-TBT:
Hi Will,
a quick fix for a bug I spotted in the pdf export. 
The "regular" text added after the 2nd dynamic label is cropped or ignored (`last_idx` was incremented instead of keeping track of the regex end position)

Should I make a post on image.sc to mention the bug and the steps to fix it? 
Replacing the latest script with this one should suffice.


will-moore:
Tested a label of `X: [x.unit] Y: [y.unit]thistextwillbecut` - without this PR I get truncated text and this PR fixes the bug 👍 
I think this can be included as a bug-fix in the next release of Figure.
If anyone reports it and needs a fix sooner then we can give them instructions, but I don't think it needs anything more urgent just yet.

Issue #442, opened 2021-07-05
Time-stamp delta-T offset


will-moore:
See https://github.com/ome/omero-guide-figure/pull/12#issuecomment-869873611

Some Images have time-stamp metadata that has a time offset from the desired time-stamp labels.
Need an option for adding or removing a fixed amount of time before formatting time-stamp labels.

This could be stored in seconds or milliseconds, but the user might want to choose the value in hrs, mins or secs etc?

cc @abhamacher

Issue #441, opened 2021-07-05
time-stamp decimal formatting


will-moore:
From https://github.com/ome/omero-guide-figure/pull/12#issuecomment-865827295

Currently the time-stamp options are:

```
Time (milliseconds)
Time (seconds)
Time (mins)
Time (mins:secs)
Time (hrs:mins:secs)
Time (hrs:mins)
```

Maybe if you choose one of those, and the text box then says [time-milliseconds] we could show an info message like:
To show decimal places, add ".0", e.g. "[time-milliseconds.00]" for 2 decimal places.

cc @abhamacher

PR #481, opened 2022-09-13
Time reference and decimal places

Tom-TBT:
Hello again,
with this PR I want to solve two issues that can be handled in the dynamic labels:
- Specifying a frame index to serve as the "zero" timepoint https://github.com/ome/omero-figure/issues/442
- Specifying the decimal precision for time, that I then also found relevant for x, y, z, width and height https://github.com/ome/omero-figure/issues/441

A picture is worth a thousand words:
image

The decimal precision and offset (the "reference frame") are passed as named parameters inside the label:
- `[time.seconds; precision=2]` -> parameters are specified after a `;`, in the format `param_name=value`
- `[time.seconds; offset=3; precision=2]` -> the order of the parameter is not important
- `[time.seconds ; offset = 3 ;precision= 2  ]` -> Spaces are not affecting the parsing
- `[x.unit; offset=3; precision=2]` -> Here the offset is not relevant for `x.unit`: it is ignored. The same happens with unknown parameters or invalid formats (no parameter name, no value, no `=` sign).

Finally, I added tests for the labels like @jburel suggested last time. I know that they should be run on github to test that nothing is crashing during the PDF conversion, but how would you run them locally to check visually that the output is the one we expect?

Cheers,
Tom

#########################################

EDIT: The following is kept for reference about the discussion in this PR, but the format that was ultimately picked is described above.

UI-wise, here is what I chose:

- I specify the decimal precision as an integer instead of a series of '0'. The serie of '0' raises question about the implementation (should I count the characters, the zeros, what should it do when there is a mix). Also it would mean that the decimal precision can be 0 only when nothing is specified, but we already had 2 as the standard for x/y/z/width/height

- decimal precision comes after the format specification (secs, ms, mins, ...). Luckily the default for time/x/y.... are indexes, so having decimal precision for them doesn't make sense. Decimal precision is **always** in 3rd position, never 2nd:
  - [time.5] -> decimal precision of 5 for the index. No one needs that and the behavior is that the format isn't recognized
  - [time.index.5] -> this also doesn't make sense but still displays the index correctly (ignoring decimal places)
  - [time.seconds.2] -> Time in seconds with decimal precision of 2.
  - [x.unit.1] -> Position in X units (like µm) with decimal precision of 1.
 
- The frame reference for the time is given at the end, after a `-`. The time of that frame is subtracted from the other timepoints. Being the last parameter, it works both with and without decimal precision:
  - [t.secs.2-3] -> Decimal precision of 2 with reference frame n°3
  - [t.secs-3] -> Reference frame n°3
  - [t.secs-0] -> 0 is too low, reference frame not understood, ignoring the reference
  - [t-3] -> Here I decided not to implement it so it just shows the index ignoring the reference
  - [x.unit-3] -> Substracting 3rd pixel x-position? I also decided not to implement it. I don't see a use for it


will-moore:
The failed run X status of this PR seems to be because the test export script is failing at `scale = panel['pixel_size_x']` (KeyError) probably because the image has no pixel size metadata.

Actually, I can reproduce the error by adding a label `X: [x.unit]` to an Image panel which has no pixel size (e.g. a png).
So it makes sense to do `scale = panel.get('pixel_size_x')` and then handle `scale is None`.

Otherwise, testing all looks good 👍

Tom-TBT:
The script failed there yes, but I wouldn't do the change in `Figure_To_Pdf.py`. I think that the test is wrong. The JSON looks like an old version! (pixel_size_x seems introduced in V1, looking at `version_transform()` in `figure_model.js`)

With the version transform, all pdf generated from now on should have the new JSON, right? (except if someone forgot to refresh his browser for quite some time now)

So I updated the JSON in the test (and reorganized it to make it look like the model in `figure_model.js`). If I'm right, there are also values added from `panel_model` (rotation, zoom, ...)

Let me know what you think about my JSON update, then let's see if the test passes this time 🤞 

PS: Working on it, I spotted another bug in `Figure_To_Pdf`: label with `[z.unit]` on 2D images made it crash (z_symbol is None)
I saved myself the effort of making another PR and fix the bug here. Is that ok for you?

Tom-TBT:
This time it comes from my JSON generation. I tried to get `pixel_size_x` similar to what is done in `views.img_data_json`, but while the imaging loading in OMERO.figure doesn't crash when the pixel size is `NOT SET`, here it complains because `px` is None (or similar).
 
When pixel size is `NOT SET`, the panel's JSON (talking about a normal figure) doesn't contain `pixel_size_x`
-> testing on it the `[x.unit]` label, it displays `NaN µm` and the Pdf export crashes!

I think that I should handle it the same way as for Z and T, displaying `0` by default 
(and without adding  `pixel_size_x` to the JSON, so that we keep the warning in the UI 
image
)

Tests are useful :D

omero_figure/templates/figure/index.html
```diff
@@ -581,14 +581,19 @@ <h4>Dynamic properties</h4>
...
                             <th>Time with frame n°3 as the "zero"</th>
-                            <td>Time (ms): [time.milliseconds-3]</td>
-                            <td>Time (ms): -58 ms</td>
+                            <td>Time: [time.milliseconds-3]</td>
+                            <td>Time: -58 ms</td>
+                        </tr>
+                        <tr>
+                            <th>Time with decimal precision (2) and "zero" (3)</th>
+                            <td>Time: [time.seconds.2-3]</td>
+                            <td>Time: -0.058 s</td>
```
will-moore, line 596: "decimal precision (2)" should be `Time: -0.06 s`?

Tom-TBT: oh yes indeed

omero_figure/templates/figure/index.html
```diff
@@ -581,14 +581,19 @@ <h4>Dynamic properties</h4>
...
                         <tr>
-                            <th>Time with decimal precision of 2</th>
-                            <td>Time (s): [time.seconds.2]</td>
-                            <td>Time (s): 0.93 s</td>
+                            <th>Width with decimal precision of 3</th>
+                            <td>Width: [width.unit.3]</td>
+                            <td>Width: 4.450 µm</td>
                         </tr>
                         <tr>
                             <th>Time with frame n°3 as the "zero"</th>
```
will-moore, line 589: I'm not sure that "zero" is the best way to explain this? Maybe "Label shows time relative to the 3rd frame". Or "Time difference between current and 3rd frames"?

omero_figure/templates/figure/index.html
```diff
@@ -581,14 +581,19 @@ <h4>Dynamic properties</h4>
...
-                            <td>Time (ms): [time.milliseconds-3]</td>
-                            <td>Time (ms): -58 ms</td>
+                            <td>Time: [time.milliseconds-3]</td>
+                            <td>Time: -58 ms</td>
+                        </tr>
+                        <tr>
+                            <th>Time with decimal precision (2) and "zero" (3)</th>
```
will-moore, line 594: "Time with decimal precision 2, relative to the 3rd frame"

src/js/models/panel_model.js
```diff
@@ -368,6 +368,8 @@
             y_size = this.get('pixel_size_y');
             z_size = this.get('pixel_size_z');
             z_size = z_size?z_size:0
+            x_size = x_size?x_size:0
+            y_size = y_size?y_size:0
```
will-moore, line 372: Minor point: I find that kinda hard to read without spaces. Could you use `y_size = y_size ? y_size : 0;`
Elsewhere I've started to use `Prettier` for code formatting to handle these cases, but not here yet... thx.

will-moore:
Testing the figure to Pdf with missing `pixel_size_x` is fixed now 👍 .
Just a couple of wording suggestions above.

omero_figure/templates/figure/index.html
Tom-TBT: "Time relative to the 3rd frame" sounds great

src/js/models/panel_model.js
Tom-TBT: yes no problem. I'll go refactor it where I find it

docs/figure_file_format.rst
```diff
@@ -129,6 +129,15 @@ Optional settings for each panel::
             "size": "12",
             "position": "topright",
             "color": "FFFFFF"
+        },
+        {
+            // for 'time', 'x', 'y', 'width' and 'height', decimal precision
+            // can be specified in 3rd position (here 2)
+            // 'time' can be relative to a frame, specified after a '-' sign (here 3)
+            "text": "Time (s): [time.secs.2-3]",
```
joshmoore, line 137: Just as a heads up, my inclination is that this format won't be readily understandable by users. Sorry if I missed it, but was there any discussion of the cost/benefit trade-off on having a larger but more self-documenting format

joshmoore: Brainstorming:
 - `time[3].{unit:.2f}` (more python-y)
 - `[time.unit;offset=3;precision=2]` (more readable)

Tom-TBT: `[time.unit;offset=3;precision=2]`   I like that one. It's readable, won't be much effort to implement and we don't need to worry about the order of parameter

will-moore:
After discussing @joshmoore and others this morning, we felt that something like this (which uses css-style formatting) would be easier to remember, more human readable and extensible: `[time.unit; offset:3; precision:2]`.
That was my web-based preference, but Josh also suggested `[time.unit; offset=3; precision=2]` or you could go for JSON style `[time.unit, offset:3, precision:2]`.

joshmoore:
> you could go for JSON style [time.unit, offset:3, precision:2].

Only slightly marred by there being no quotes :)

Tom-TBT:
I will be on holiday for the next two weeks. 
Maybe best not to apply this change right now, so we have time to think about it. 

But I like those "self-documenting formats"
thanks a lot for the ideas!

Tom-TBT:
Hello Moore²,
I ended up picking this style: `[time.unit; offset=3; precision=2]`. The `:` sign is already used for the `h:m:s` format so I wanted to avoid the confusion (though it's not a problem for the parsing, as `;` is parsed first).

I also updated the tests, the tip window, and the figure_file_format documentation accordingly.
Let me know!

will-moore:
Sorry, not looked yet - I'm away till Wednesday...
Please feel free to ping me again after then.

src/js/views/panel_view.js
```diff
@@ -196,23 +196,42 @@
...
-                        } else if (['image', 'dataset'].includes(expr[0])){
-                            label_value = self.model.get_name_label_text(expr[0], expr[1] ? expr[1] : "name");
+
+                        // Label parsing in three steps:
+                        //   - split label type.format from other parameters (;)
+                        //   - split label type and format (.)
+                        //   - grab other parameters (key=value)
+                        expr = match[0].slice(1,-1).split(";");
+                        prop_nf = expr[0].trim().split(".")
+                        param_dict = {};
```
will-moore, line 206: these variable declarations need `var` or `let`

src/js/views/panel_view.js
```diff
@@ -196,23 +196,42 @@
...
+                        });
+
+                        var label_value = "";
+                        if (['time', 't'].includes(prop_nf[0])) {
+                            format = prop_nf[1] ? prop_nf[1] : "index";
+                            precision = param_dict["precision"] !== undefined ? param_dict["precision"] : 0; // decimal places default to 0
+                            label_value = self.model.get_time_label_text(format, param_dict["offset"], precision);
+                        } else if (['image', 'dataset'].includes(prop_nf[0])){
+                            format = prop_nf[1] ? prop_nf[1] : "name";
+                            label_value = self.model.get_name_label_text(prop_nf[0], format);
```
will-moore, line 221: Need to check that all these variable declarations have `var` too, or they end up as global variables.

will-moore:
Ah - I blindly tried `[time.unit; offset=3; precision=2]` for a time label (from the comment above) and this didn't get recognised... I realised after a while that `time.unit` is not valid.

Could you update the PR description with the new format (this can also serve as a reference)

will-moore:
Functionality testing looks good. Labels are rendered the same in JavaScript and Python.
Just a couple of minor comments above.

will-moore:
Thanks Tom - some others on the team are away just now, but I when they're back we'll discuss if there's any outstanding issues/feedback. Cheers

pwalczysko:
Other than https://github.com/ome/omero-figure/pull/481#issuecomment-1287137105, this PR LGTM

Tom-TBT:
Hi Petr,
it looks to me that markdown is responsible for it. 

Looking a bit around I found this: https://gist.github.com/dupuy/1855764#block-quotations
quote: "Markdown treats four leading spaces as a code example"

So I don't know if bug or feature since we can now add code to the labels!!! ... Well maybe more like a bug, that's a weird way of attaching code to images

We could remove the leading spaces, but that could destroy someone's label alignment workflow. Let me see if I find a way of keeping the spaces without doing some markdown with it. Tell me if you would have that in this PR or in a separate one.

Tom-TBT:
Well maybe that becomes hard to handle, I tested many from that tutorial https://www.markdownguide.org/basic-syntax/ and many works:
image

will-moore:
Created an issue for the formatting behaviour the @pwalczysko reported: https://github.com/ome/omero-figure/issues/486
@jburel you happy to merge this?

omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
```diff
@@ -1087,34 +1087,36 @@ def get_crop_region(self, panel):
 
         return {'x': cropx, 'y': cropy, 'width': tile_w, 'height': tile_h}
 
-    def get_time_label_text(self, delta_t, format):
+    def get_time_label_text(self, delta_t, format, dec_prec):
```
jburel, line 1090: A default value should be given to ``dec_prec`` e.g. ``dec_prec = -1`` so we do not change the behaviour of the method

omero_figure/templates/figure/index.html
```diff
@@ -580,6 +580,21 @@ <h4>Dynamic properties</h4>
                             <td>*Height*: **[height.pixel]**</td>
                             <td><i>Height</i>: <strong>300</strong></td>
                         </tr>
+                        <tr>
+                            <th>Width with decimal precision of 3</th>
+                            <td>Width: [width.unit; precision=3]</td>
+                            <td>Width: 4.450 µm</td>
```
jburel, line 586: ``µm`` this should probably be encoded e.g. ``&#181; ``

src/js/models/panel_model.js
```diff
@@ -277,50 +277,58 @@
...
             var pad = function(digit) {
                 var d = digit + "";
                 return d.length === 1 ? ("0"+d) : d;
             };
             var theT = this.get('theT'),
                 deltaT = this.get('deltaT')[theT] || 0,
-                isNegative = (deltaT < 0),
                 text = "", h, m, s;
+
+            if (ref_idx){
```
jburel, line 297:
```suggestion
            if (ref_idx) {
```

Tom-TBT:
Thank you Jean-Marie for the review, I made the changes

jburel:
@Tom-TBT sorry for the delay. Thanks for making the adjustments

will-moore:
Thanks @Tom-TBT - this is now released in omero-figure 5.1.0

Issue #314, opened 2018-10-25
No ":" in default figure name


joshmoore:
see:
 * https://github.com/ome/omero-figure/blob/6c1c258c96670265a53610dda62f832b2efb458c/src/js/models/figure_model.js#L421
 * https://twitter.com/totti0223/status/1055364434632171520

> i do not like to have colon included as a default file name candidate.  (regular expression things)

will-moore:
Fixed in https://github.com/ome/omero-figure/pull/497

PR #497, opened 2023-02-08
Default figure filename to YYYY-MM-DD_hh-mm-ss

Tom-TBT:
Hello Will,
tiny PR to replace `:` by `-` for the default figure filename, to solve https://github.com/ome/omero-figure/issues/314

That also includes the padding of date and time with zeros

If you would rather keep the `:`, I can revert to that format but keep the padding (otherwise reject the PR :D).

Side note, could you also close those solved issues:
- https://github.com/ome/omero-figure/issues/318  -> solved by https://github.com/ome/omero-figure/pull/472
- https://github.com/ome/omero-figure/issues/441 -> solved by https://github.com/ome/omero-figure/pull/481
- https://github.com/ome/omero-figure/issues/442 -> solved by https://github.com/ome/omero-figure/pull/481

Best,
Tom

### DIFF
--- a/src/js/models/figure_model.js
+++ b/src/js/models/figure_model.js
@@ -550,9 +550,14 @@
         },
 
         getDefaultFigureName: function() {
+            const padL = (nr, len = 2, chr = `0`) => `${nr}`.padStart(2, chr);
             var d = new Date(),
-                dt = d.getFullYear() + "-" + (d.getMonth()+1) + "-" +d.getDate(),
-                tm = d.getHours() + ":" + d.getMinutes() + ":" + d.getSeconds();
+                dt = [d.getFullYear(),
+                      padL(d.getMonth()+1),
+                      padL(d.getDate())].join('-'),
+                tm = [padL(d.getHours()),
+                      padL(d.getMinutes()),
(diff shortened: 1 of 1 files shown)


src/js/models/figure_model.js
```diff
@@ -550,9 +550,14 @@
         },
 
         getDefaultFigureName: function() {
+            const padL = (nr, len = 2, chr = `0`) => `${nr}`.padStart(2, chr);
```
will-moore, line 553: `len` argument isn't used in `padStart()`, but you only ever use the first argument anyway, so you could remove the `len` and `chr` args?

will-moore:
Issues closed, thx.

If you include e.g. "Fixes #314" in your PRs, that should auto-close the issue when the PR is merged.

src/js/models/figure_model.js
Tom-TBT: Sorry for that, corrected in the last commit

Tom-TBT:
Thanks for the tip about auto close issues, will add the right keyword from now on.

will-moore (approved):
👍

PR #504, opened 2023-03-23
Add option to adjust Scalebar width

### LINKED ISSUE #503: Scalebar line thickness
See https://forum.image.sc/t/scalebar-width-adjustement/78827

"Is it possible to set the scalebar width ? For some figure, it appears a bit thick and we would like to thin it a bit, with a drop-down menu (like for label font size) or simply manually."

JulianHn:
Hey all,

This is a first draft to Fix #503 

I have added an additional input field to the scalebar_form panel and reordered the inputs to make space.
Adjusting the scalebar in the panel_view is done.

image

image

Known Issue: Site load show Empty string passed to getElementbyId(). This will also be returned multiple times when the scalebar is shown / hidden. Unclear where this is coming from

RFC: Is that approach in principle okay? If yes, I will try to track done that getElementbyId() thing and take a look on what needs to be modified in the export routines.

// Julian

### DIFF
--- a/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
+++ b/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
@@ -2217,10 +2217,8 @@ class TiffExport(FigureExport):
         y2 = scale_to_export_dpi(y2)
         width = scale_to_export_dpi(width)
 
-        for l in range(width):
-            draw.line([(x, y), (x2, y2)], fill=rgb)
-            y += 1
-            y2 += 1
+        for l in range(-width // 2, width // 2):
+            draw.line([(x, y+l), (x2, y2+l)], fill=rgb)
 
     def draw_temp_label(self, text, fontsize, rgb):
         """Returns a new PIL image with text. Handles html."""
--- a/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
+++ b/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
@@ -1493,8 +1493,9 @@ class FigureExport(object):
                 pass
 
             # For 'bottom' scalebar, put label above
+            # 5 is an empirically determined offset that "works"
             if 'bottom' in position:
-                ly = ly - font_size
+                ly = ly - font_size - 5
             else:
                 ly = ly + 5
 
@@ -2297,7 +2298,6 @@ class TiffExport(FigureExport):
     def draw_text(self, text, x, y, fontsize, rgb, align="center"):
         """ Add text to the current figure page """
         x = scale_to_export_dpi(x)
-        y = y - 5       # seems to help, but would be nice to fix this!
         y = scale_to_export_dpi(y)
         fontsize = scale_to_export_dpi(fontsize)
 
--- a/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
+++ b/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
@@ -901,7 +901,7 @@ class FigureExport(object):
         if self.zip_folder_name is not None:
             full_name = os.path.join(self.zip_folder_name, full_name)
 
-        while(os.path.exists(full_name)):
+        while os.path.exists(full_name):
             index += 1
             full_name = "%s_page_%02d.%s" % (name, index, fext)
             if self.zip_folder_name is not None:
@@ -1477,7 +1477,8 @@ class FigureExport(object):
         else:
             lx_end = lx - canvas_length
 
-        self.draw_scalebar_line(lx, ly, lx_end, ly, sb["height"], (red, green, blue))
+        self.draw_scalebar_line(lx, ly, lx_end, ly, sb["height"],
+                                (red, green, blue))
 
         if 'show_label' in sb and sb['show_label']:
             symbol = u"\u00B5m"
--- a/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
+++ b/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
@@ -1434,19 +1434,20 @@ class FigureExport(object):
         position = 'position' in sb and sb['position'] or 'bottomright'
         align = 'left'
 
+        half_height = sb['height'] // 2
         if position == 'topleft':
             lx = x + spacer
-            ly = y + spacer
+            ly = y + spacer + half_height
         elif position == 'topright':
             lx = x + width - spacer
-            ly = y + spacer
+            ly = y + spacer + half_height
             align = "right"
         elif position == 'bottomleft':
             lx = x + spacer
-            ly = y + height - spacer
+            ly = y + height - spacer - half_height
         elif position == 'bottomright':
             lx = x + width - spacer
-            ly = y + height - spacer
+            ly = y + height - spacer - half_height
             align = "right"
 
         pixel_size_x = panel['pixel_size_x']
@@ -1476,7 +1477,7 @@ class FigureExport(object):
         else:
             lx_end = lx - canvas_length
 
-        self.draw_scalebar_line(lx, ly, lx_end, ly, 3, (red, green, blue))
+        self.draw_scalebar_line(lx, ly, lx_end, ly, sb["height"], (red, green, blue))
 
         if 'show_label' in sb and sb['show_label']:
             symbol = u"\u00B5m"
@@ -1498,7 +1499,10 @@ class FigureExport(object):
                 ly = ly + 5
 
             self.draw_text(
-                label, (lx + lx_end)/2, ly, font_size, (red, green, blue),
+                label, (lx + lx_end)/2,
+                ly + ((-1 if position in ["bottomleft", "bottomright"]
+                       else 1) * half_height),
+                font_size, (red, green, blue),
                 align="center")
 
     def is_big_image(self, image):
--- a/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
+++ b/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
@@ -1477,7 +1477,7 @@ class FigureExport(object):
         else:
             lx_end = lx - canvas_length
 
-        self.draw_scalebar_line(lx, ly, lx_end, ly, sb["height"],
+        self.draw_scalebar_line(lx, ly, lx_end, ly, sb.get("height", 3),
                                 (red, green, blue))
 
         if 'show_label' in sb and sb['show_label']:
--- a/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
+++ b/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
@@ -1434,7 +1434,7 @@ class FigureExport(object):
         position = 'position' in sb and sb['position'] or 'bottomright'
         align = 'left'
 
-        half_height = sb['height'] // 2
+        half_height = sb.get('height', 3) // 2
         if position == 'topleft':
             lx = x + spacer
             ly = y + spacer + half_height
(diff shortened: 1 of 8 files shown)


JulianHn:
PDF export seems to be working as well with modifactions to `Figure_to_Pdf.py`:
[Figure_2023-3-23_13-52-0.pdf](https://github.com/ome/omero-figure/files/11051663/Figure_2023-3-23_13-52-0.pdf)
[Figure_2023-3-23_13-51-44.pdf](https://github.com/ome/omero-figure/files/11051671/Figure_2023-3-23_13-51-44.pdf)

I haven't fully understand why the TIFF export breaks with the same parameters ... more investigation to do

JulianHn:
And with [1cd3f56](https://github.com/ome/omero-figure/pull/504/commits/1cd3f560fa71070851134b05c688289810ee8a86) exporting tiffs now also works on a first glance:

image

image

JulianHn:
After some further research:
The `Empty string passed to getElementById()` warning  is also present in current stable omero-figure and occurs exclusively on Firefox. This seems to be a known issue with Firefox and Bootstrap Forms and Dropdown menus and the general consensus seems to be that it is not an actual problem.

E.g:
https://community.wix.com/velo/forum/coding-with-velo/text-input-for-forms-generates-empty-string-passed-to-getelementbyid-error-on-load
https://bugzilla.mozilla.org/show_bug.cgi?id=1333427

The second link and some other StackOverflow threads often involve data-targets pointing to `#` as the culprit. I am wondering if the <a href="#"> parts of the dropdown menus for units and font size are responsible?

will-moore (changes requested):
@JulianHn, thanks for the PR - you've done well to figure all that out!

Just a couple of small points...

The buttons and form elements are slightly bigger than before. You need to add the `btn-sm` and `input-sm` classes to them.

Before/after comparison:

Screenshot 2023-03-24 at 10 10 56


Minor point: I wonder if you could get the form to fit on 2 rows instead of 3? If it's too painful don't worry. But if you could get the show/hide button to float right (so as to align with the other forms) that would be great (`pull-right` class).

It seems that adding an extra text input to the form has disabled the "submit on Enter" behaviour for the length input (and the height input). So that when you enter a length or hight and hit Enter, nothing happens. Previously I was getting the "submit" event for free! Just looked this up at https://www.tjvantoll.com/2013/01/01/enter-should-submit-forms-stop-messing-with-that/

So we need to add that behaviour explicitly. This will do it...

```
--- a/src/js/views/scalebar_form_view.js
+++ b/src/js/views/scalebar_form_view.js
@@ -27,6 +27,14 @@ var ScalebarFormView = Backbone.View.extend({
         "click .pixel_size_display": "edit_pixel_size",
         "keypress .pixel_size_input"  : "enter_pixel_size",
         "blur .pixel_size_input"  : "save_pixel_size",
+        "keyup input[type='text']"  : "handle_keyup",
+    },
+
+    handle_keyup: function (event) {
+        // If Enter key - submit form...
+        if (event.which == 13) {
+            this.update_scalebar();
+        }
     },
 ```

The PDF and TIFF export look great. However I noticed an existing issue that could do with improving (while we're here): In the PDF (right) the label is probably a bit too close to the scalebar: it looks OK on the TIFF. Could you give it a bit more space?

Screenshot 2023-03-24 at 10 35 05

Cheers!

**Comment by will-moore** [comment]
Oh, and maybe remove the `?` from `Label?`. I think a checkbox is implicitly boolean.

**Comment by JulianHn** [comment]
Hey @will-moore 
thanks for the review. 
Fitting the everything in two rows was no problem, my original thought was separating "label-logic" from "bar-logic".

image

All other requested changes are implemented and Enter submission is now working again by applying your patch and fixing a css selector.

The label offset in PDF is now consistent with the one in TIFF, I just used the 5 offset that was already implemented there for both now.

PDF:
image
image


TIFF:
image
image


I have undrafted the PR, as I think it should now have all features and I have done a bit more UI testing without encountering more issues.

// Julian

**Comment by sbesson** [comment]
The PR inclusion will be conditional to its status. I suspect the issue came from the default requirement to approve GitHub workflows for first-time contributors. Looking at the actions logs, the initial run four days ago was marked as `required action` and it looks like  you approved it and the GitHub workflow ran successfully 2 days ago.
Screenshot 2023-03-28 at 11 02 49

Regarding the commit, note that https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-plugins-push/ did not run at all on March 27th so this PR did not have a change to be included yesterday. The nightly CI run seems to have successfully included it into the merge build as expected - see https://github.com/snoopycrimecop/omero-figure/commit/0948e32c8c1e5d47bcb094d3dc0c543405e5e14f

**Comment by will-moore** [comment]
@sbesson Ah - thanks, sorry I'd missed that the CI didn't run yesterday.

File: src/js/views/scalebar_form_view.js [hunk 1]
```diff
@@ -27,7 +27,15 @@ var ScalebarFormView = Backbone.View.extend({
         "click .pixel_size_display": "edit_pixel_size",
         "keypress .pixel_size_input"  : "enter_pixel_size",
         "blur .pixel_size_input"  : "save_pixel_size",
-    },
+        //"keyup input[type='text']"  : "handle_keyup",
```
will-moore, line 30: I don't know how this got commented-out, but it's still needed!

will-moore:
@sbesson the last snoopy merge commit (last night) https://github.com/snoopycrimecop/omero-figure/commit/1cd70919c15ddf29b069074c16da72fb6e430bc8 again failed to merge this PR because `action_required` again, and I see above that I have to `Approve and run` again, because 1 commit got added?!

I would have expected to only need to approve once?

will-moore:
Hmmm - now "Approve and run" gives me a 500 error at https://github.com/ome/omero-figure/pull/504/run_workflows so will try again later...

will-moore (approved):

will-moore:
@sbesson I'm still getting this PR excluded - just now: https://github.com/snoopycrimecop/omero-figure/commit/c6c6834fae45f9e56d3e032e98240416a250b8e4 this time with `(status: startup_failure)`. 
I don't see anything in the logs at https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-plugins-push/1475/console

sbesson:
Screenshot 2023-03-30 at 09 22 51

Looks like the latest workflow run hit some upstream service unavailability issue

will-moore:
@sbesson I can't actually work out how to get to that view of the jobs. 
If I follow the "details" link from the job I get to https://github.com/ome/omero-figure/actions/runs/4542452995/jobs/8034996020?pr=504
but I can't figure out how to get to https://github.com/ome/omero-figure/actions/runs/4542452995/ to see the error (except by editing the URL)!?

That shows:
```
Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/checkout@v2.
For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.
```
cc @jburel

sbesson:
From https://github.com/ome/omero-figure/actions/runs/4542452995/jobs/8034996020?pr=504, can you click on the `Latest #3` drop-down then select `Latest attempt #3`?

will-moore:
Thank's to @jburel and https://github.com/ome/omero-figure/pull/505 it looks like the build is passing now...

JulianHn:
Just to be sure:
Do I just add the single line to the `JSON` optional settings specification or is there some documentation of all the options I am missing?

Regarding your NB: I actually do not really see much "how to use" in the current README in general. It's more focused on "how to sysadmin / develop" and the "how to use" is not really linked at all. Should the README be extended in general or do you think a prominent link to the readthedocs would be sufficient.

// Julian

will-moore:
@JulianHn - apologies this has taken a while to get it deployed on our testing server (I was just testing locally before) but it's there now...

I just noticed that when I opened a previously-saved figure, the scalebars don't show up on the figure because the scalebar JSON has no `height`.
And when I try to generate a PDF it fails:

```
  File "./script", line 1963, in add_panels_to_page
    self.draw_scalebar(panel, pil_img.size[0], page)
  File "./script", line 1437, in draw_scalebar
    half_height = sb['height'] // 2
```
This should simply use `half_height = sb.get('height', 3) // 2` here and anywhere else it reads the height in the script.

Probably the best way to handle this in the app is to increment the VERSION at https://github.com/ome/omero-figure/blob/master/src/js/models/figure_model.js#L4 to `7`.

Then add a block below this section, similar to the handling of addition of scalebar units to version 5:
https://github.com/ome/omero-figure/blob/master/src/js/models/figure_model.js#L189
to add a default `scalebar.height` (of 3 I think it was).

Cheers

pwalczysko:
Additionally what @will-moore wrote, a minor comment - See screenshots below, we have now on the same ``Labels`` tab in the OMERO.figure two ways to denote the font size, the tooltip says something else for existing Labels and something else for the newly added ``Height``. And the tooltip itself has also different size in the ``Height`` case (smaller) and in the ``Labels`` case (bigger) - see screenshots below.

Existing Labels functionality, denoted as ``Font size``
Screenshot 2023-04-08 at 14 30 26

The newly added ``Height`` denoted as ``Label size``
Screenshot 2023-04-08 at 14 30 17

pwalczysko:
@JulianHn Sorry, deleted my comment about ``font`` and ``label`` discrepancy, as it was not introduced by this PR, and created an issue https://github.com/ome/omero-figure/issues/507 instead.

JulianHn:
Hey @will-moore,

sorry for the delay - I was on vacation / conference the last 1.5 weeks.
For some reason it did not cross my mind to save and reload a figure while testing.
I'll take a look and add the necessary changes to this PR.

@pwalczysko: While it might not have been introduced here, I can add the necessary changes in this PR, if you prefer the layout in the `Labels` case. I'll track #507

// Julian

pwalczysko:
> @pwalczysko: While it might not have been introduced here, I can add the necessary changes in this PR, if you prefer the layout in the `Labels` case. I'll track #507

Thank you very much for this generous suggestion. Yes, we discussed with @will-moore : 
Let us go for `Labels` case, this means the tooltip says `Font size` everywhere (scalebar and labels) and is the same size and style as in the present state in `Labels`. Definitely not a blocker of course.

will-moore:
@sbesson it seems I need to "Approve and run" this PR after every commit to avoid it being excluded by merge-ci?
Any way to make this a bit easier?

sbesson:
This is definitely the expected  behavior and I don't think there is a built-in feature to automatically approve every new commit - see https://github.blog/2021-04-22-github-actions-update-helping-maintainers-combat-bad-actors/ for background. Once the PR of a new contributor is accepted, the approval should no longer be mandatory for future contributions so this is a one-off requirement but it will apply to every new commit of the first PR.

omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
```diff
@@ -1476,7 +1477,8 @@ def draw_scalebar(self, panel, region_width, page):
         else:
             lx_end = lx - canvas_length
 
-        self.draw_scalebar_line(lx, ly, lx_end, ly, 3, (red, green, blue))
+        self.draw_scalebar_line(lx, ly, lx_end, ly, sb["height"],
```
will-moore, line 1480: I missed this one earlier - Also needs to be `sb.get("height", 3)`

will-moore:
The loading of pre-existing Figure files is working now. 👍 
Just one more comment above for the python script.
On second thoughts, I don't think #507 has to be fixed in this PR (can follow-up later).

JulianHn:
I also missed it :)
Your suggested patch in #507 is applied to this tree and tested on Chrome and Firefox on Linux Mint 21.
Now all the menus have Bootstrap tooltips setup, the only inconsistency is that the tooltip for the Font Size in the scalebar form has a linebreak for me. But I am guessing this depends on the screen size etc. and I don't think it makes the whole form look inconsistent.

will-moore:
Thanks @JulianHn - looks good 👍

pwalczysko:
Screenshot 2023-05-03 at 09 48 59

Is the Font Size widget completely missing now from the Labels ^^^ ? Looks like it...

pwalczysko:
Sorry, https://github.com/ome/omero-figure/pull/504#issuecomment-1532663856 is not precise - the Font Size box appears when the Label checkbox is checked. All good here.

Screenshot 2023-05-03 at 09 55 44

pwalczysko (approved):

Issue #428, opened 2021-03-11
Zoom over 1000%


will-moore:
See https://forum.image.sc/t/omero-iviewer-zoom-limits/39220/3

Want to support zooming over 1000%, but don't want the slider to be too insensitive at lower zoom values.
One solution might be to allow users to edit the Zoom value as a Text field (as you can do with channel rendering sliders).

This is similar to #371 (text field for rotation slider).

will-moore:
Also requested at https://forum.image.sc/t/couple-of-suggestions-about-omero-figure/72675 to manually enter the Zoom value.

Rdornier:
Hello @will-moore, 

Do you think that the manual way to enter the zoom value will be integrated soon ?
I have a project with whole slide images and the value of the zoom is quite critical. Using the slider, it is not possible to achieve the value we want.

I'm also happy to contribute a bit to omero-figure but when I browse the repo, I was a but lost between the different files (json, html) and I do not really know where to modify things.

Thanks,
Rémy.

Rdornier:
Ok, thanks for the infos.
I tried the workaround ; it works but it's not convienent as users should also do this as well.
But at least, they have a way to set the same zoom level.

I'm happy to see the new features in that update :)

PR #517, opened 2023-09-15
Manual zoom level 

Rdornier:
Hello, 

I try to make to necessary changes to add manually the zoom level (fix part of #428).
However, I struggle a lot to test the changes in my dev env (i.e. changes in the code are not always taken into account when I test them).
So, feel free to test and to give me feedback.

Rémy.

### DIFF
--- a/src/js/views/right_panel_view.js
+++ b/src/js/views/right_panel_view.js
@@ -623,12 +623,7 @@
                 self.listenTo(m,
                     'change:channels change:theZ change:theT',
                     self.rerender_image_change);
-              //  self.listenTo(m, "change:$vp_zoom_value", self.handle_user_zoom)/
             });
-
-            //document.getElementById("vp_zoom_value").addEventListener("keyup", (event) => {self.handle_user_zoom} );*/
-
-           // $('#vp_zoom_value').mouseup(handle_user_zoom)
 
             $("#vp_zoom_slider").slider({
                 // NB: these values are updated on render()
@@ -637,8 +632,7 @@
                 value: self.zoom_avg,
                 slide: function(event, ui) {
                     self.update_img_css(ui.value, 0, 0);
-                   // $('#vp_zoom_value').val(ui.value);
-                    console.log("je passe par ici")
+                    $('#vp_zoom_value').val(ui.value);
                 },
                 stop: function( event, ui ) {
                     self.zoom_avg = ui.value;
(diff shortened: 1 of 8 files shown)


will-moore:
Hi @Rdornier - see https://github.com/Rdornier/omero-figure/pull/1 for some fixes

Rdornier:
Hi @will-moore 

Many thanks for your corrective input ; it perfectly works 👍 !
I've added few more lines to update the image in the figure. 
I tested it and it work fine for me.

src/js/views/right_panel_view.js
```diff
@@ -702,6 +708,44 @@
...
+
+            // update the slider position
+            $("#vp_zoom_slider").slider({'value': value});
+
+            // update the preview image
+            this.update_img_css(value, 0, 0);
+
+            // update the figure image
+            this.zoom_avg = value;
+            var to_save = {'zoom': value};
```
will-moore, line 737: @Rdornier Just testing this, and I realise that the `value` never gets converted from a string to an Integer.
JavaScript works OK with the string as it casts it on the fly for e.g. `value < minVal` etc.
But when we save it to Figure JSON model, it gets saved as a string (since we don't have any validation of the model).
This means that when you subsequently select multiple panels, the average zoom values aren't calculated properly. E.g. `"100" + "200" = "100200"` !

Rdornier: @will-moore Ok, I've just added a commit to convert the value coming from the UI from string to int (in the listener function and the slider implementation)

src/js/views/right_panel_view.js
```diff
@@ -631,11 +631,11 @@
...
                 value: self.zoom_avg,
                 slide: function(event, ui) {
-                    self.update_img_css(ui.value, 0, 0);
-                    $('#vp_zoom_value').val(ui.value);
+                    self.update_img_css(parseInt(ui.value), 0, 0);
+                    $('#vp_zoom_value').val(parseInt(ui.value));
                 },
                 stop: function( event, ui ) {
-                    self.zoom_avg = ui.value;
+                    self.zoom_avg = parseInt(ui.value);
```
will-moore, line 638: You don't actually need `parseInt()` for the `slide` and `stop` functions since the `ui.value` will already be a Number.

Rdornier: Ok, I remove it from these two functions

will-moore:
Looking good - I think it could do with a CSS tweak. Trying this:

```
$ git diff
index 332b657..ad108bb 100644
@@ -233,6 +233,11 @@
         height: 14px;
         border-radius: 5px;
     }
+    #vp_zoom_value {
+        float: right;
+        width: 50px;
+        margin-top: 5px;
+    }
     .toggle_channel {
```

gives a nicer layout:

Screenshot 2023-09-28 at 16 36 07

will-moore (approved):
Thanks!

PR #520, opened 2023-10-03
Align right bottom

Rdornier:
Add a button to align images to the right and to the bottom

### DIFF
--- a/omero_figure/templates/figure/index.html
+++ b/omero_figure/templates/figure/index.html
@@ -865,7 +865,7 @@
                         <button type="button" class="aleft btn btn-default btn-sm" title="Align Left" disabled="disabled">
                             <span class="glyphicon glyphicon-align-left"></span>
                         </button>
-                        <button type="button" class="aleft btn btn-default btn-sm" title="Align Right" disabled="disabled">
+                        <button type="button" class="aright btn btn-default btn-sm" title="Align Right" disabled="disabled">
                             <span class="glyphicon glyphicon-align-right"></span>
                         </button>
                         <button type="button" class="agrid btn btn-default btn-sm" title="Align To Grid" disabled="disabled">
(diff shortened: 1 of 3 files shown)


will-moore:
This works OK if all the panels are the same width or height, but if they aren't then it should probably still align the right or bottom edges. So you need to combine `x` and `width` to find the right edge for each, and the max right-edge. Then subtract the width of each panel from that value when setting the `x`. And same logic for bottom alignment.

Rdornier:
Arf... I was doing my tests on multiple copies of the same image and did see that.
Thansk for reporting this bug !

will-moore:
This is working well now.

There's a minor issue with the extra space being taken up on the tool bar: The extra buttons are contributing to an existing bug where the toolbar is spreading into space used by the figure title, which is an issue if you have a longer title or narrow browser page:

Screenshot 2023-10-09 at 18 16 29

The previous issue of a gap to the right of the toolbar can be fixed by moving it to the right.
Then we still need to reduce the space used by the title.
With these changes...

```
index 332b657..f5b049c 100644
@@ -123,7 +123,7 @@
         line-height: 20px;
         padding: 12px;
         position: absolute;
-        right: 450px;
+        right: 520px;
         text-align: center;
         height: 100%;
     }
@@ -705,6 +705,10 @@
     .navbar>div.container {
         max-width: 100%;
     }
+    #alignment-toolbars {
+        padding-right: 0;
+        margin-right: -15px;
+    }
     /** Add my class to boost size of font icons **/
     .icon-buttons {
         margin-right: 20px;
```

Looks a bit better:

Screenshot 2023-10-09 at 18 29 21

Rdornier:
It can indeed be annoying; thank you for your meticulous tests !

will-moore (approved):

PR #525, opened 2023-11-01
Annotation inheritance support

Tom-TBT:
This PR is at a testing stage and relies on a changed function (`api_annotations`) from omero-web, inside a [parallel PR](https://github.com/ome/omero-web/pull/508).

Inherited annotations are decompressed, and the ann.link.parent is changed to the inheritor object. Like so, all further handling of the annotation is the same. No difference if the annotation was on the current image or on a parent object.

The implementation also checks that the query's result has the inherited annotations.


Tom-TBT:
Thank you Will, I could reproduce the error and your solution fixed it.

will-moore (approved):
Works now, thanks LGTM 👍

PR #528, opened 2023-11-09
Labels from ALL KVP 

Rdornier:
This should fix issue #501 
See PR #529 for last commit 73f6477 on underscore escape

Issue #486, opened 2022-10-25
Label markdown formatting does too much


will-moore:
See https://github.com/ome/omero-figure/pull/481#issuecomment-1287137105

The markdown formatter that is used for italics and bold formatting of labels, will also format other markup.
E.g. if the label starts with 4 white-spaces, it will be formatted as code (see comment above) or if starts with a ` - ` it will add a bullet point.

Probably makes sense to simply strip all white-space from user-entered text.
We could also strip ` - ` and then add it back after formatting with markdown, in case user actually wants " - my label".

will-moore:
Also, user may actually want `text_with_underscores` not to be formatted as italics. See https://forum.image.sc/t/bugs-with-labels-in-omero-figures/78829.
We could give user the option to disable/enable markdown formatting for a label, but then useful italics etc wouldn't be possible. Better info on how to `\_escape_` markdown might help too.

Rdornier:
@will-moore can be closed

PR #529, opened 2023-11-10
Escape underscores on template labels

Rdornier:
Fix part of issue #486, related to 

I didn't change much because I think the markdown reading need to be preserved (see PR https://github.com/ome/omero-figure/pull/481#issuecomment-1288570316).

For each label added from a template (tags, channel name and key-values only ; image and dataset name are correctly handled), the underscore is escaped with a backslah ``replaceAll("_", "\\_")``. This enables a correct display of the label in omero.figure and it is compatible with ``Figure-to-pdf.py`` (``_`` and ``\_`` are identically interpretated in the script).

I've also integrated this feature in the PR #528.

### DIFF
--- a/src/js/views/labels_from_maps_modal.js
+++ b/src/js/views/labels_from_maps_modal.js
@@ -85,7 +85,7 @@ var LabelFromMapsModal = Backbone.View.extend({
             if (imageKeyValues[iid]) {
                 var labels = imageKeyValues[iid].map(function(value){
                     return {
-                        'text': includeKey ? (key.replaceAll("_","\\_") + ': ' + value.replaceAll("_","\\_")) : value.replaceAll("_","\\_"),
+                        'text': includeKey ? (value[0].replaceAll("_","\\_") + ': ' + value[1].replaceAll("_","\\_")) : value[1].replaceAll("_","\\_"),
                         'size': labelSize,
                         'position': labelPosition,
                         'color': labelColor,
--- a/src/js/views/labels_from_maps_modal.js
+++ b/src/js/views/labels_from_maps_modal.js
@@ -80,7 +80,7 @@ var LabelFromMapsModal = Backbone.View.extend({
             if (imageValues[iid]) {
                 var labels = imageValues[iid].map(function(value){
                     return {
-                        'text': includeKey ? (key + ': ' + value) : value,
+                        'text': includeKey ? (key.replaceAll("_","\\_") + ': ' + value.replaceAll("_","\\_")) : value.replaceAll("_","\\_"),
                         'size': labelSize,
                         'position': labelPosition,
                         'color': labelColor,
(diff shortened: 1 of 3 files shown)


snoopycrimecop:
Conflicting PR. Removed from build [OMERO-plugins-push#311](https://snoopycrimecop-ci.openmicroscopy.org/jenkins/job/OMERO-plugins-push/311/). See the [console output](https://snoopycrimecop-ci.openmicroscopy.org/jenkins/job/OMERO-plugins-push/311/consoleText) for more details.
Possible conflicts:
  - PR #528 Rdornier 'Labels from ALL KVP '
    - src/js/views/labels_from_maps_modal.js

--conflicts

snoopycrimecop:
Conflicting PR. Removed from build [OMERO-plugins-push#312](https://snoopycrimecop-ci.openmicroscopy.org/jenkins/job/OMERO-plugins-push/312/). See the [console output](https://snoopycrimecop-ci.openmicroscopy.org/jenkins/job/OMERO-plugins-push/312/consoleText) for more details.
Possible conflicts:
  - PR #528 Rdornier 'Labels from ALL KVP '
    - src/js/views/labels_from_maps_modal.js

--conflicts

src/js/views/labels_from_maps_modal.js
```diff
@@ -85,7 +85,7 @@ var LabelFromMapsModal = Backbone.View.extend({
             if (imageKeyValues[iid]) {
                 var labels = imageKeyValues[iid].map(function(value){
                     return {
-                        'text': includeKey ? (value[0].replaceAll("_","\\_") + ': ' + value[1].replaceAll("_","\\_")) : value[1].replaceAll("_","\\_"),
+                        'text': includeKey ? (key.replaceAll("_","\\_") + ': ' + value.replaceAll("_","\\_")) : value.replaceAll("_","\\_"),
```
will-moore, line 88: I think that in fixing those merge conflicts you've got an error here. `value` is actually a list of `[key, value]`. It actually looks like you don't need any change to this line, since you've already got the escaping logic there?

Rdornier: True there was an error here. I didn't notice it before. Thanks

Rdornier:
I have a question regarding the documentation in the "Tips". It is written that to make a text italic, you need to use one start \*italic\* but it is also working with underscores \_italic\_. Should we add the two options in the tips ? 
And should we also inicate how to escape the underscores to not be interpreted as markdown formatting ?

Rdornier:
I see your point...
Preferentially, I would keep the current implementation because it is quite simple and consistant. However, if you really see an issue with escaping all underscores, then I can implement the counting. We can also ask the question to @jburel 

I'm also thinking about a different solution to use the markdown. By default, nothing is interpretated as markdown. If the user introduce ``md:`` at the beginning, then the label is markdown compatible.
I don't know if this solution could be good / feasible ; it just comes out of the mind. It will however introduce a major change and we should open a need issue / PR to discuss about it.

will-moore:
OK, agreed. 👍

Issue #502, opened 2023-03-22
Duplicate Labels can't be edited independently


will-moore:
See https://forum.image.sc/t/bugs-with-labels-in-omero-figures/78829

Best option might be to avoid creation of duplicates, particularly from e.g Tag -> Labels dialog etc.

Rdornier:
@will-moore Issue can be closed

PR #530, opened 2023-11-10
Duplicated labels removal

Rdornier:
Fix issue #502 by removing duplicate labels.
1. If a new label is added, the duplicated label is filtered before adding it to the figure
2. If an existing label is edited, the duplicated label is filtered after adding it to the figure.

### DIFF
--- a/src/js/models/panel_model.js
+++ b/src/js/models/panel_model.js
@@ -252,7 +252,6 @@
                 oldLabKeys.push(lbl_key);
                 labs.push( $.extend(true, {}, lbl));
             }
-
             // ... then add new labels ...
             for (var j=0; j<labels.length; j++) {
                 var lbl = labels[j];
@@ -262,7 +261,6 @@
                     labs.push( $.extend(true, {}, lbl));
                 }
             }
-
             // ... so that we get the changed event triggering OK
             this.save('labels', labs);
         },
@@ -435,6 +433,7 @@
         // labels_map is {labelKey: {size:s, text:t, position:p, color:c}} or {labelKey: false} to delete
         // where labelKey specifies the label to edit. "l.text + '_' + l.size + '_' + l.color + '_' + l.position"
         edit_labels: function(labels_map) {
+            
             var oldLabs = this.get('labels');
             // Need to clone the list of labels...
             var labs = [],
--- a/src/js/models/panel_model.js
+++ b/src/js/models/panel_model.js
@@ -243,24 +243,24 @@
             var oldLabs = this.get("labels");
             // Need to clone the list of labels...
             var labs = [];
-            var newLabKey = [];
-            for (var i=0; i<labels.length; i++) {
-                newLabKey.push(this.get_label_key(labels[i]));
-            }
+            var oldLabKeys = [];
 
+            // Keep old labels unchanged...
             for (var i=0; i<oldLabs.length; i++) {
-                lbl = oldLabs[i];
-                lbl_key = this.get_label_key(lbl);
-                // for existing label that matches...
-                if (!newLabKey.includes(lbl_key)) {
-                    // otherwise leave un-edited
-                    labs.push( $.extend(true, {}, lbl));
-                }
+                var lbl = oldLabs[i];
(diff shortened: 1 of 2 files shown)


src/js/views/right_panel_view.js
```diff
@@ -524,15 +522,25 @@
...
+                    filteredLbls.push(lbl)
+                })
+                
                 json.inner_template = self.inner_template;
                 html += self.template(json);
             });
+           
+            this.models.forEach(function(m){
+                m.save('labels', filteredLbls);
+            })
```
will-moore, line 542: I'm not sure what the changes in `right_panel_view.js` are doing (are they needed)?
But this seems to be breaking "undo" functionality since with every re-render you are saving a change to the model.

Rdornier: It is to handle any label edition. For example, if you have two labels ``aa``, one green, one blue, both at the same size and position, then if you turn the green one to blue, only one blue label is remaining. But obviously, if it breaks the undo, I have to do it differently

src/js/models/panel_model.js
```diff
@@ -243,13 +243,26 @@
...
+
             for (var i=0; i<oldLabs.length; i++) {
-                labs.push( $.extend(true, {}, oldLabs[i]) );
+                lbl = oldLabs[i];
+                lbl_key = this.get_label_key(lbl);
+                // for existing label that matches...
+                if (!newLabKey.includes(lbl_key)) {
+                    // otherwise leave un-edited
+                    labs.push( $.extend(true, {}, lbl));
+                }
```
will-moore, line 258: If I add a new label and it already exists, it removes the existing one and adds the new one, which is slightly unexpected. E.g. If I add labels, `1`, `2`, `3`, `4`, and then add another duplicate label `2`, then I now have labels
`1`, `3`, `4`, `2`, whereas I'd expect the attempt to add a duplicate should have no effect.

will-moore:
The `Undo` functionality is working again, thanks. I wonder if you could clean-up the unnecessary whitespace diff lines left over since it increases the chance that you'll get a merge conflict with other PRs.

src/js/models/panel_model.js
Rdornier: Ok, I have corrected it.

src/js/models/panel_model.js
```diff
@@ -444,8 +455,20 @@
...
+            // filter unique keys only to remove duplicates
+            var filtered_lbls_map = {}
+            for (var i=0; i<labs.length; i++) {
+                filtered_lbls_map[this.get_label_key(labs[i])] = i
+            }
+        
+            var filtered_lbls = [] 
+            _.each(Object.values(filtered_lbls_map), function(idx){
+                filtered_lbls.push(labs[idx])
+            })
```
will-moore, line 468: This is working fine for me now, thanks.
I don't want to sound too picky, but I had a slightly hard time understanding the code, and I'm also not sure that using `Object.values` will always preserve the order of the items (although it seems to be working).
I just wonder, since you're filtering, if it's clearer to use the `filter` function like this?

```
            // filter unique keys only to remove duplicates
            var keys = new Set(labs.map(lbl => this.get_label_key(lbl)));
            var filtered_lbls = labs.filter(lbl => keys.has(this.get_label_key(lbl)));
```

Apologies: I understand that this may not be a "better" way for everyone, and I don't want to make code review and PR submission any more time-consuming than it has to be. So please feel free to ignore this suggestion if it doesn't feel better for you!

Rdornier: > I don't want to sound too picky,

Don't worry, you have much more experience than me on Javascript !

However, I tested your changes but duplicates are not always removed. If  `labs` contains duplicate labels, then duplicates are not removed.

Rdornier: The snippet now works correctly. 
It is definitely more concise and I think it could be understandable quite easily (filtering based on the index of the first occurrence of the value)
So it should be ok 👍

Rdornier: It also removes the dirty ``Object.values()``, which I don't really like as well

PR #537, opened 2024-01-03
Update infra

jburel:
Adjust code and fix test following the update of the docker images used in omero-test-infra
Note that this PR depends on https://github.com/ome/omero-test-infra/pull/66
**DO NOT MERGE** until the omero-test-infra is merged


jburel:
> So I presume that the changes to `jburel/omero-test-infra` will be merged into the mainline in due course, then we can revert to using the mainline here?
Exactly
> 
> There weren't any changes needed in copying the PDF export script from the current branch?

No

PR #541, opened 2024-01-25
Don't swap width/height when cropping page

### LINKED ISSUE #539: Crop fails if page/panel is in horizontal orientation
Hi Will,

I often use the "Crop page around panels" option in the File/Page Setup menu, but there is one case where it unfortunately fails.

The workflow where it happens is:

1. Set the page size to A4 **horizontal** orientation
2. Create a panel of images
3. Use "Crop page around panels" 

To illustrate it, here's a very basic example **before cropping**:

image

**After cropping** you get: 

image

The workflow works fine, when the original page is in vertical orientation.

Would be really nice if this could be fixed at some point for the horizontal orientation.

Thank you, Anna

will-moore:
Fixes #539 

To test:

 - In page layout, choose "horizontal" page orientation
 - Arrange some panels in a non-square layout (so you can tell the difference between width and height)
 - Now in page layout again, choose "crop" to panels
 - Cropped page should match the panels (not swap width/height as in bug on issue above)

### DIFF
--- a/src/js/views/modal_views.js
+++ b/src/js/views/modal_views.js
@@ -156,7 +156,7 @@
                 h_mm = Math.round(h_pixels * 25.4 / dpi);
             }
 
-            if (orientation == 'horizontal' && size != 'mm') {
+            if (orientation == 'horizontal' && size != 'mm' && size != 'crop') {
                 var tmp = w_mm; w_mm = h_mm; h_mm = tmp;
                 tmp = w_pixels; w_pixels = h_pixels; h_pixels = tmp;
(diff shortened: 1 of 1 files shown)


pwalczysko:
tested on ome-training-3 (without this PR) and merge-ci (with this PR).

without this PR, the problem manifests
Screenshot 2024-02-16 at 16 57 36

with this PR, the problem is gone
Screenshot 2024-02-16 at 16 57 28

pwalczysko (approved):

PR #543, opened 2024-02-11
Plate well labels

### LINKED ISSUE #540: Additional labels for plate images
Hi Will,

I produce a lot of plate data and am a happy user of omero.figure to create panels of these images. These panels tend to get big and it's really easy to make c&p mistakes. To avoid this, it would be really helpful to automatically add the well name and field name as a label to each image. At the moment "Image Name" and "Dataset Name" can be added automatically, but neither well, field, nor plate name. I would really appreciate if these options could be added too. The field name could be a simple number, or the style it is already shown in omero.web:

image

Possible ideas:
image

Would be both fine and making at least my life definitely easier. Any other implementation would be fine too, as long as the label is populated automatically. Would that be possible? 

Of course, sometimes the field is not important and the well name is sufficient, so I would suggest to keep it variable to the user if which labels are used.

Thanks in advance,
Anna

will-moore:
Fixes #540.

 - Fix `[wellsample.index]` label to be `[field.index]` and to be 1-based instead of 0-based.
 - Handle edit of label -> plate.name etc as well as *add* label.
 - Add 'Screen' option everywhere

### DIFF
--- a/src/js/views/panel_view.js
+++ b/src/js/views/panel_view.js
@@ -219,11 +219,12 @@
                             format = prop_nf[1] ? prop_nf[1] : "index";
                             precision = param_dict["precision"] !== undefined ? param_dict["precision"] : 0; // decimal places default to 0
                             label_value = self.model.get_time_label_text(format, param_dict["offset"], precision);
-                        } else if (['image', 'dataset'].includes(prop_nf[0])){
+                        } else if (['image', 'dataset', 'screen', 'plate', 'well', 'wellsample'].includes(prop_nf[0])){
                             format = prop_nf[1] ? prop_nf[1] : "name";
                             label_value = self.model.get_name_label_text(prop_nf[0], format);
+                            console.log("render_label label_value", label_value, prop_nf);
                             //Escape the underscore for markdown
-                            label_value = label_value.replaceAll("_", "\\_");
+                            label_value = ("" + label_value).replaceAll("_", "\\_");
                         } else if (['x', 'y', 'z', 'width', 'height', 'w', 'h', 'rotation', 'rot'].includes(prop_nf[0])){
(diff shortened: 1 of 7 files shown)


src/templates/labels_form_inner_template.html
Tom-TBT, line 1: `<li><a href="#" data-label="[well.label] Field#[field.index]">Well & Fieild</a></li>`
Mispell of Field

src/js/views/panel_view.js
```diff
@@ -219,11 +219,12 @@
                             format = prop_nf[1] ? prop_nf[1] : "index";
                             precision = param_dict["precision"] !== undefined ? param_dict["precision"] : 0; // decimal places default to 0
                             label_value = self.model.get_time_label_text(format, param_dict["offset"], precision);
-                        } else if (['image', 'dataset'].includes(prop_nf[0])){
+                        } else if (['image', 'dataset', 'screen', 'plate', 'well', 'wellsample'].includes(prop_nf[0])){
```
Tom-TBT, line 222: Could you add `acquisition` to the list of accepted objects? It's already part of the parent in the JSON response.
 
And/or the alias `run` with a ternary condition inside the if:
`prop_nf[0] = prop_nf[0] == "run" ? "acquisition" : prop_nf[0]`

src/templates/labels_form_inner_template.html
Tom-TBT: Also `[field.index]` is not functional. 
Either needs to be changed to `[wellsample.index]` or could be given an alias in the JS (similar to the comment on run/acquisition)

will-moore:
Thanks for the feedback and suggestions @Tom-TBT.
But I probably won't get back to finishing this off until after the vite work is a bit more complete.

PR #548, opened 2024-03-12
Fix cropping of panel labels and float usage for shape labels

will-moore:
Since we updated to use `x, y, x2, y2 = font.getbbox(text)` to get the height of text, the labels that are pasted on to the temporary canvas of that height are cropped at the bottom. We have removed the margin by using `height = x2 - x` but since the pasting at `x = 0` actually pastes *with* the margin (so the canvas is not big enough).

If we set the temporary canvas to grey for each label for debugging, we can see this effect more clearly:

Screenshot 2024-03-12 at 13 09 47

 In order to paste right at the top of the temp canvas (without the margin), we need to set the x coordinate to be negative margin:

This also positions the text the same as in the original figure:

Screenshot 2024-03-12 at 13 19 51

Original figure:

Screenshot 2024-03-12 at 13 21 33

NB: I also tested the export of *shape* labels (shown in 2nd screenshot above) which didn't need a fix to the `font.getbbox()` logic, but did need some casting of floats to ints. Shape labels aren't supported by vanilla OMERO.figure - I think they were added to the export script by Glencoe, so the only way to test that is to add `"text": "some text"` to the Shape JSON. cc @knabar 

To test:

 - Add labels to a Figure (including with markdown italics and bold) and export as TIFF - check that they are not cropped in the TIFF

### DIFF
--- a/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
+++ b/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
@@ -564,7 +564,7 @@ class ShapeToPilExport(ShapeExport):
             return
         r, g, b, a = self.get_rgba_int(shape['strokeColor'])
         # bump up alpha a bit to make text more readable
-        rgba = (r, g, b, 128 + a / 2)
+        rgba = (r, g, b, int(128 + a / 2))
         font_name = "FreeSans.ttf"
         from omero.gateway import THISPATH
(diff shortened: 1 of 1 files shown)


pwalczysko:
compare the same figure on demo (without this pr)
Screenshot 2024-04-12 at 14 04 17

and on merge-ci (with this PR)
Screenshot 2024-04-12 at 14 04 11


lgtm

pwalczysko (approved):

PR #527, opened 2023-11-09
Zoom label

Rdornier:
Adding the zoom level in the label list

PR #549, opened 2024-04-08
Inset feature

will-moore:
Fixes #319 

This replaces the tricky "create Inset" workflow (copy Rectangle and paste as Crop region or vice versa) with a single button click, and the Inset panel and Rectangle should stay in sync when either are updated.

To test:

- Select one or more panels and in the right-tab, under `ROIs`, click the "Create Inset" button
- This will create a new Inset Panel and a corresponding Rectangle:

Screenshot 2024-04-26 at 11 06 04

- When you Pan, Zoom, rotate or Resize the Inset Panel, the Rectangle will update in sync (NB: we previously handled rotation when copying crop region and pasting as ROI by creating a Polygon instead of rotated Rectangle)
- When you edit the Rectangle (via the ROI dialog) the Inset Panel will update accordingly
- If you copy and paste the Inset panel to create e.g. split-view panels, each of these panels is "linked" to the original Rectangle (updating any one of them will update the Rectangle and vice-versa)
- If you copy and paste the parent panel that contains the Rectangle, both Rectangles in these panels are linked to the Inset Panel(s)
- Save the Figure, then re-open (can simply refresh the page) and check that the Rectangle and Inset panels are still kept in sync.

PR #550, opened 2024-04-08
Don't update zoom to undefined on setId

will-moore:
Fixes a bug introduced in #527: On `Edit ID` the panel zoom is updated to `undefined` since the incoming image data has no zoom info so `data.zoom` is always undefined.

This results in rendering errors - e.g. Zoom slider `0` and other `NaN` values below:

Screenshot 2024-04-08 at 17 49 39

To test:

 - Zoom a panel, then `Edit ID` to replace image with another.
 - The zoom should remain unchanged and after de-selecting and re-selecting the panel (to trigger update of right panels) the zoom controls, viewport x, y, width, height etc should be rendered correctly.

cc @Rdornier

### DIFF
--- a/src/js/models/panel_model.js
+++ b/src/js/models/panel_model.js
@@ -103,7 +103,6 @@
                 'pixel_size_x_unit': data.pixel_size_x_unit,
                 'pixel_size_z_unit': data.pixel_size_z_unit,
                 'deltaT': data.deltaT,
-                'zoom':data.zoom,
             };
 
             // theT and theZ are not changed unless we have to...


will-moore:
@jburel - Any idea what's failing in the above job? Seems that `npm` isn't installed.

Same error on #549 (and I assume on all other PRs when they next try to run checks):

```
+ python setup.py sdist
running sdist
npm install
error: command 'npm' failed: No such file or directory
```

Thanks

Rdornier:
I can indeed reproduce the bug. 
I didn't thought about testing this feature when I added this feature.
Thanks for reporting and fixing it.

pwalczysko:
Reproduced on demo (without this PR). See it fixed on merge-ci.

pwalczysko (approved):

will-moore:
Released in omero-figure 6.2.1

PR #551, opened 2024-04-08
bump nodejs

jburel:
Bump nodejs and use dnf
This should fix the build failure reported by @will-moore


will-moore (approved):
Looks good and build is green, thanks

PR #559, opened 2024-04-29
Vite 3.2.10

will-moore:
This fixes and replaces #557.


will-moore:
Instead, pushed fix to #557 itself

PR #561, opened 2024-05-03
Bugfixes may2024

will-moore:
Fixes for any bugs introduced with the vite.js upgrade:

 - 1) Channel sliders: for small ranges (e.g. float data where brightest pixel is less than 100) we use a smaller step for the slider and we format the value to float when displaying in the start/end text boxes. We also have logic to prevent the 'end' slider value going less than the 'start' slider. In this bug, I wasn't using the full pixel range to check for using small step and float formatting - therefore the test image (see fake image below) with pixel range of `-32768.00` - `72` was using the float formatting etc. Also, the formatting to string step was happening before we compared `start < end` values, so we were comparing string with number which gave a wrong result (so that when sliding the 'end' value, it would get reset to the same as the start value.

`testProjection&sizeX=1952&sizeY=1952&sizeZ=73&sizeC=4&pixelType=int16.fake`

### DIFF
--- a/src/js/views/channel_slider_view.js
+++ b/src/js/views/channel_slider_view.js
@@ -219,8 +219,8 @@ var ChannelSliderView = Backbone.View.extend({
         let $target = $(event.target);
         let value = parseFloat(event.target.value);
         let max = $target.attr('max');
+        let min = $target.attr('min');
         let chIndex = $target.data('idx');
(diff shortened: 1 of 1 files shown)


pwalczysko:
Tested on merge-ci

- without this fix, the slider on the max side is vanishing when slided to the left, and the max value is wrongly formatted
- with this fix, the slider is working fine and the max value is as expected.

lgtm

PR #563, opened 2024-06-05
Use path instead of re_path for all URLs

will-moore:
I was reminded in prep for recent Web workshops that the Django recommended way of defining URLs is with `path()` function rather than `re_path()` and has been so since Django 3.2 at least.
E.g. see https://docs.djangoproject.com/en/4.2/topics/http/urls/

This updates OMERO.figure to use `path()` as a first step. Can think about omero-web and other apps too...

Expect no change in functionality.

To test all URLs...

 - [x] Go To /figure/open - open a new figure (list Files and thumbnails OK)
 - [x] Go To /figure/new - Add a new image to figure
 - [x] Save the Figure
 - [x] Test rendering of Big images
 - [x] Delete a figure
 - [x] Add an image with timestamps to figure, check timestamps are loaded `timestamps/?image=id
 - [x] Move figure to another group
 - [x] figure/z_scale/?image=id
 - [x] figure/roiCount/<int:image_id>/
 - [x] figure/recover/
 - [x] figure/roiRectangles/image_id
 - [x] figure/unit_conversion/<int:value>/<slug:from_unit>/<slug:to_unit>/',

### DIFF
--- a/omero_figure/urls.py
+++ b/omero_figure/urls.py
@@ -18,7 +18,7 @@
 
 from omeroweb.webgateway import views as webgateway_views
 from . import views
-from django.urls import re_path, path
+from django.urls import path
 
 
 urlpatterns = [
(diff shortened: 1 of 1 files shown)


jburel:
This PR should probably be included in the upcoming 7.0.0

will-moore:
@jburel I have just checked all the features/URLs above which covers all URLs on merge-ci.

PR #575, opened 2024-07-03
Label sorting with drag and drop

Tom-TBT:
Hello Will,
here's a PR to sort labels with a drag-and-drop. The cursor changes when hovering the labels, and moving the labels up or down, the order of the labels will be updated.

For the tests I conducted it works well. This handles:
* multiple panels (all labels equal, all different, mix of both)
* multiple label position (one set of positions is updated at a time)

This is the work from my new student Tehmina who is helping me with changes I'd like to make. @MinaEnayat

### DIFF
--- a/src/js/views/right_panel_view.js
+++ b/src/js/views/right_panel_view.js
@@ -548,52 +548,78 @@
             return false;
         },
 
-        initializeSortable: function() {
+        initialize_sortable: function() {
             var self = this;
-            var el = self.$el.find('.labels-container').get(0);
-    
-            // Check if the element exists   
-            if (!el) {         
-                console.warn('Sortable element not found.');         
-                return;       
+            var containers = self.$el.find('.labels-container');
+
+            // Check if there are any elements
+            if (containers.length === 0) {
+                return;
             }
-    
-            // Destroy existing Sortable instance if any
-            if (self.sortable) {        
-                self.sortable.destroy();        
-                self.sortable = null;      
-            }      
-                  
-            self.sortable = new Sortable(el, {         
-                animation: 150,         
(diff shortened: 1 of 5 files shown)


src/js/views/right_panel_view.js
```diff
@@ -548,6 +549,79 @@
...
+            var self = this;
+            var containers = self.$el.find('.labels-container');
+
+            // Check if there are any elements
+            if (containers.length === 0) {
+                return;
+            }
+
+            // Destroy existing Sortable instances if any
+            if (self.sortables) {
```
will-moore, line 562: PR looks great and works well!

The only question I have is about the `destroy()` here... 
If I create a sortable, then select a different panel in the figure then this `SelectedPanelsLabelsView` gets lost without destroying the selectable(s) that were created.
I don't know if that really matters - probably not. Does it hurt if we don't clear destroy them? E.g. use up memory etc??
If we do want to make sure they all get destroyed, I wonder if a more global list of sortables (instead of just `this` component) or possibly try to handle the removal of the component and destroy them there. E.g. see https://github.com/ome/omero-figure/blob/64566080f813fb5d70a77c3cd1a141bf30eaddba/src/js/views/right_panel_view.js#L756 (which was originally created to handle the destruction of jQuery sliders - now I don't use jQuery sliders, so it does less now).

Tom-TBT: I followed the clear example and now destroy the sortables when the panel is deselected. Thanks for spotting that

will-moore (approved):
Thanks for the fix - testing locally looks good and functional testing on merge-ci is behaving as expected 👍

will-moore:
This is now released in omero-figure 7.1.0

Issue #506, opened 2023-04-04
Weird translation if spanning image preview after cropping and rotating image


Rdornier:
## Bug report

**Describe the bug**
When spanning an image, first cropped, then rotated, on the preview panel, the image preview is translated and does not correspond anymore to what is really displayed.

**To Reproduce**
1. Open OMERO.figure
2. Import an image
3. Draw an ROI on the image
4. Crop the image around the ROI
5. Rotate the cropped image
6. Span the image in the preview tab

**Expected behavior**
Having the right preview when spanning the image on the preview tab

**Screenshots**
![OMERO figure_Bug1](https://user-images.githubusercontent.com/64911638/229802819-9cae53af-6408-4583-8cf7-d07b11d926ef.gif)

**Version :**
 - OMERO.figure version: 5.1.0

Rdornier:
Hello, 

I had a closer look to this issue and can update a bit.
- The weird translation seems to only happen for big images (I tried with a small image and it translates correctly)
- It also happens if you don't rotate the image (i.e. only crop it)
- If you click on the zoom slider (don't necessarly slide it), the weird translation disappear.

So, it seems that something is not / wrongly saved when the crop is performed. The wrong values are not affected when the user spans in the preview (so, the weird translation occurs each time you span). However, the slider acts as a reset button whenever you click / slide it and then everything is fine.

I had a look to the code but I couldn't find anything that would explain this behavor for now.
I'll keep searching how to fix this bug but any inputs are welcome !

Rémy.

PR #580, opened 2024-07-19
Weird translation

Rdornier:
Fixes issue #506


will-moore:
Thanks for the PR fix. I'm away till August so won't be able to look more just yet...

will-moore:
Thanks for tracking down a fix for this.
The change you've made here has some issues when you select 2 panels that have different zoom values. In that case the zoom average is a very different value from the existing zoom value of each panel, so that `m.set(zoom)` actually changes the figure quite a lot by updating those panels.
Actually, it shouldn't ever be necessary to call `model.set()` in a `render()` method, since that will update the figure while the user is just viewing the figure (instead of making any desired changes).

However, the instructions for reproducing the bug and the change you've made (that did fix it nicely for a single panel) did allow me to work out the core issue, which is that the `cropToRectangle()` sets the zoom to a float (instead of an integer) and this isn't handled well by the panning behaviour.

One possible fix that works is to make sure that `cropToRectangle()` saves the zoom as an Integer:

```
--- a/src/js/models/panel_model.js
+++ b/src/js/models/panel_model.js
@@ -610,7 +610,7 @@
             // zoom to correct percentage
             var xPercent = this.get('orig_width') / coords.width,
                 yPercent = this.get('orig_height') / coords.height,
-                zoom = Math.min(xPercent, yPercent) * 100;
+                zoom = parseInt(Math.round(Math.min(xPercent, yPercent) * 100));
```

However, I don't think that we should rely on zoom always being an Integer, so I looked for the bug in panning and tracked the fix down to `model.get_vp_big_image_css()`. Since the `zoom` value that is passed into that function is always an Integer (it comes from zoom_avg), when `this.get('zoom')` is a float then `zooming` is True, which it shouldn't be when we're panning. If we always compare the *difference* it fixes the issue:

```
--- a/src/js/models/panel_model.js
+++ b/src/js/models/panel_model.js
@@ -790,7 +790,7 @@
             // Used for static rendering, as well as during zoom, panning, panel resizing
             // and panel re-shaping (stretch/squash).
 
-            var zooming = zoom !== this.get('zoom');
+            var zooming = Math.abs(zoom - this.get('zoom')) > 1;
             var panning = (x !== undefined && y!== undefined);
```

will-moore:
Thanks - merging.
NB: you've got the same change to .gitignore in #581 but you can resolve that when fixing merge conflicts there...

PR #581, opened 2024-07-22
Fix panel & label zoom value

Rdornier:
fix 2 small bugs regarding the zoom
- update the panel when copy-pasting the zoom from one image to another.
- cast to integer the value of the zoom when it is displayed as label

Before
image

After
image


will-moore:
Hi @Rdornier, apologies for the delayed response.
Can you describe the bug you're fixing a bit more? What exactly is "copy-pasting the zoom from one image to another"?

Without this PR, when I use the `Copy` and `Paste` buttons to copy the crop region from one panel to another, it seems to work OK for me.
However, to test the formatting of the label I created `"zoom": 267.444` by creating a figure from hand-tweaked JSON. Copy and Paste of crop region didn't give the same zoom, probably due to rounding errors when converting zoom -> rectangle and then rectangle -> zoom:

After `Copy` from the top image and `Paste` to the lower one (Zoom slider and value have updated):

Screenshot 2024-08-05 at 17 30 51

Rdornier:
Hi @will-moore 

Thanks for checking.
The bug was, when I used the `Copy` and `Paste` buttons to copy the crop region between panels, the viewport and the slider didn't update "on the fly". I need to deselect the panel and select it again to see the changes.

However, this issue happened on my developement environment but I've just tried on the production server and this issue doesn't happend at all (both have the same version of omero.figure 7.0.0) ! So, forget about the panel update. I reversed the code to original.

Regarding the label formatting and different zooms, I fixed the rounding errors. It should be good now. I keep the zoom value as a float value everywhere except when it is displayed, in which case it is rounded.

will-moore:
Looks good, thanks.
Can you merge in the master branch and resolve the conflicts so that this branch is mergeable? Thx

PR #583, opened 2024-08-23
Add Hilo Checkbox for LUT Toggling in Channel Slider View

MinaEnayat:
Hello,
I've implemented a new feature to add a Hilo checkbox to the channel slider view. 

This checkbox allows users to rapidly toggle the Hilo LU across all channels. 

When the checkbox is checked, the Hilo LUT (hilo.lut) is applied to all channels, and when unchecked, the original colors are restored.

Please review the changes, and feel free to provide any feedback or suggestions.

Mina

@Tom-TBT

### DIFF
--- a/src/js/models/panel_model.js
+++ b/src/js/models/panel_model.js
@@ -27,9 +27,6 @@
             pixel_size_x_unit: 'MICROMETER',
             rotation_symbol: '\xB0',
             max_export_dpi: 1000,
-            hilo_enabled: false,
-            hilo_original_colors: [],
-
             // 'export_dpi' optional value to resample panel on export
             // model includes 'scalebar' object, e.g:
             // scalebar: {length: 10, position: 'bottomleft', color: 'FFFFFF',
--- a/src/js/models/panel_model.js
+++ b/src/js/models/panel_model.js
@@ -27,6 +27,8 @@
             pixel_size_x_unit: 'MICROMETER',
             rotation_symbol: '\xB0',
             max_export_dpi: 1000,
+            hilo_enabled: false,
+            hilo_original_colors: [],
 
             // 'export_dpi' optional value to resample panel on export
             // model includes 'scalebar' object, e.g:
--- a/src/js/views/channel_slider_view.js
+++ b/src/js/views/channel_slider_view.js
@@ -26,7 +26,7 @@ function debounce(func, wait) {
 }
 
 var ChannelSliderView = Backbone.View.extend({
-
+    
     template: _.template(channel_slider_template),
     checkboxTemplate: _.template(checkbox_template),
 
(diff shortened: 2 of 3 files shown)


will-moore:
This seems to be working OK, but it is causing a couple of things to go wrong with the Undo/Redo functionality.
I realise that there's no documentation on this, so I really need to write some guidance on how this works and things to watch out for...

The Undo/Redo code listens for any `model.set()` calls and when an attribute changes then this is added to the Undo queue. So it's really important that you only call `model.set()` when the user actually edits something. This means you must avoid calling `m.set()` when you're simply rendering the UI.
If you open a previously-saved Figure with this PR, and click around a few panels to change the selection, then you go to Undo a few times, you'll see that it's selecting the same panels. This is because on selection change, the `m.set('hilo_enabled', false);` is adding the selected panel to the Undo queue, even though there's no change in the UI.

So, I would simply ignore this step and I think the code will still work OK?

...TBC...

will-moore:
Forgot to say: Thanks Mina for your first PR!
Also, we'll need a CLA form from you - see https://ome-contributing.readthedocs.io/en/latest/cla.html. Thanks!

MinaEnayat:
Hello Will, 

Thank you for your feedback and suggestions. They were incredibly helpful, especially in loading the images and applying the necessary functionalities.

I also included your bug fix of createUndo in undo.js

Best regards,
Tehmina Enayat

will-moore:
This is working fine now, thanks.
It seems to me that when you have HiLo LUT enabled for the image panel, you really just want to have a single channel active at a time, similar to the greyscale mode that we support in the other viewers?
I can think of 2 parts of supporting this: one is easier than the other:
 - When you enable HiLo, instead of only recording the colours for each channel, you could also record which channels are active. Then, you could also turn off all channels except the first of those that was active. When you turn HiLo off, you can then revert the active/inactive status of all the channels.
 - Slightly harder is to *enforce* that only a single channel is on at a time - this would mean that when you turn one channel on (and HiLo is enabled) then you also turn the other(s) off.

Am I right in thinking this would improve the feature? If so, do you want to try it? (I can give pointers if you need)?

Tom-TBT:
Hi Will,
I implemented both. I'm pleased with having only one channel at a time. 

But about restoring the active state of the channels, I'm unsure. 
It's useful for certain cases, but other times it gets in my way. Ultimately, it's a choice so I'd like your opinion on that (@pwalczysko too).

will-moore:
@Tom-TBT @MinaEnayat this works nicely and I like it. I think it works well to revert to the previously active channels when you turn HiLo off.
The only tiny tweak I though of now, is that I don't think you need `Toggle` beside the checkbox. Just `HiLo LUT` feels like enough, since a checkbox is always used to toggle.

will-moore (approved):
Looks great, 👍

pwalczysko (approved):

PR #584, opened 2024-08-30
Link url to image

### LINKED ISSUE #514: Add the link of the OMERO.figure if images are added to the figure
# Enhencement feature

## Feature description
When an image is added to a figure (and when the figure is saved), it could be great to add the OMERO.figure link as KVP pair to all images within the figure. This feature could be particularly useful when the pdf is not yet generated to easily open the figure (and see which images are part of figures)

Mirrored from this, if an image is removed from a figure, then the corresponding KVP should be also removed.

ping @romainGuiet

Rdornier:
Fixes #514 by adding a KVP on each image used a the figure. The KVP consists on the Figure URL


will-moore:
A few flake8 errors in the build:
```
./omero_figure/views.py:373:39: E203 whitespace before ':'
./omero_figure/views.py:376:80: E501 line too long (116 > 79 characters)
./omero_figure/views.py:411:30: F821 undefined name 'rlist'
./omero_figure/views.py:411:80: E501 line too long (84 > 79 characters)
./omero_figure/views.py:414:26: F821 undefined name 'rlist'
./omero_figure/views.py:432:80: E501 line too long (101 > 79 characters)
```
NB: you should be able to use `wrap( ["some string"] )` to wrap a list

Rdornier:
Hi @will-moore 

I fixed the flake8 errors.
I also fixed another bug to support the case when a figure already exists but the kvp is not yet created.

I finally noticed that `BASE_WEBFIGURE_URL` only contains the url from `/figure/` and not from `https://omero.server/webclient`. Do you know how I can get this info ?

will-moore:
Seems to be working fine from my testing so far...
I wonder if you could move that new code into a separate function(s), to keep the `save_web_figure()` function as simple/readable as possible.

Rdornier:
Ok cool !

> I wonder if you could move that new code into a separate function(s), to keep the save_web_figure() function as simple/readable as possible.

I agree, it could be better

I also noticed that I should remove the KVP from images when the figure is deleted, which is not yet the case.

will-moore:
I think it would be better not to include the figure Name in the Key because it's tricky to handle re-naming.
E.g. when I renamed a Figure and saved it, I got a new KVP annotation created, and the old one will be forever left on the images, even when they are removed from the Figure:

Screenshot 2024-10-03 at 11 48 53

Everything else seems to be working fine (delete Figure removes KVP etc).

will-moore (approved):
Looking good, thanks.

PR #555, opened 2024-04-25
Points in shape editor

will-moore:
NB: this is on top of Vite PR #477 - needs to be merged first.

As discussed at https://github.com/ome/omero-iviewer/pull/465#issuecomment-2066771147, need to support Points in OMERO.figure.

Points were already supported in `Figure_To_Pdf.py`.

To test:

 - Add Points from OMERO
 - Draw, update (drag, change colour), copy & paste Points
 - Save Points to the Figure - open Figure to check they are correct
 - Export Figure to PDF and to TIFF

PR #587, opened 2024-09-13
Update licenses.txt

will-moore:
See https://github.com/ome/omero-figure/pull/555#issuecomment-2348843360

Update licenses.txt to add/remove libraries.

cc @jburel.


jburel (approved):
Thanks @will-moore

PR #597, opened 2024-10-10
Rounding zoom label in pdf

Rdornier:
#581 introduces a fix to round the zoom label but I forgot to also round it during PDF generation.
This PR should fix it.


will-moore (approved):
Fix looks good and works fine. 👍

PR #598, opened 2024-10-11
Query timestamps from all planes when C or Z dimensions are not filled

Tom-TBT:
Mirror fix of https://github.com/ome/omero-iviewer/pull/486
Please prioritize discussion in the omero-iviewer thread


will-moore (approved):
Similar fix to iviewer. Works fine.

PR #600, opened 2024-11-08
Label on rois

### LINKED ISSUE #457: labels on ROI
No option to add text label to ROIs in figure:

See https://forum.image.sc/t/omero-figure-roi-labelling/63659

Rdornier:
Fixes #457 

image

The text is created at the same time as the other shapes and they are linked together. It means that
- Text follows shapes motion
- Text position is adapted according the shape dimension and stroke width (no manual positionning except the positions within the drop-down menu)
- If the shape is deleted, the text is also deleted
- Shapes and text color are the same

This PR is a draft for the moment, because it is not ready yet
1.  It only works for rectangle shapes for now : have to extend to any shapes
2.  The text is not displayed when the modal is closed (but apparently, they are saved, as the number of ROIs displayed in the right panel matches the total number rectangle+text) and it is not display anymore if the modal is open again.
3. Create automatically text (e.g A,B,C...) when adding insets
4. The toolbar layout have to be revised (removing `add` button,  fixing the behavior of the text input...)

@will-moore I was wondering if you could help me on point 2 ; I really don't know why shapes don't appear on the corresponding panel. Added to the shape manager, they should be visible.  I'm probably missing something but I don't know what. I'll still work on that and keep you posted if I managed to find a solution.
Thanks !


Rdornier:
Ok, I finally found a solution for point 2. I had to rescale the exported coordinates by the zoom_fraction in the json.

Rdornier:
@will-moore I've reached the point where I think it is important for me to get your feedback (I'm not asking for a entire review ; just having a look to see if there is anything obvious I've missed and maybe if the implementation looks correct for you).

Here is a summary of what I've done so far
- Only one text label is allowed per shape
- Text labels are created and added to the shapeManager when shapes are created
- Text label color is the same as the shape stroke color
- The JSON shape includes the ID of the attached text to be able to find it within the manager.
- When insets are created, automatic text labels are also created on both ROI and panel
- When a shape is pasted, the text is replaced by an empty string

For now, everything should work ONLY with rectangle shapes, including the export PDF. 
Before extending to other shapes, I would prefer getting your feedback (easiest to modify stuff only for one kind of shape).

I still need to implement (I hope I don't forget anything)
- shape rotation handling
- export tiff
- extend text to all shapes

I'm happy to discuss on anything you think relevant and to take suggestions if you have some.
Thanks.

Rémy.

will-moore:
This seems to be mostly working OK for me. I never saw a problem with shape labels showing up in the main figure:

Screenshot 2024-11-29 at 10 18 20

The only bug I've found so far is that the Shape labels aren't getting removed from the Edit ROIs image panel when the dialog is closed and reopened - either with a new panel or with a different panel.

E.g. re-opening the same panel can give an effect like this:

Screenshot 2024-11-29 at 10 25 22

will-moore:
Build is failing with:

```
flake8.main.application   MainProcess    667 INFO     Found a total of 42 violations and reported 4
./omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py:376:5: E303 too many blank lines (3)
./omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py:391:11: E275 missing whitespace after keyword
./omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py:394:13: E275 missing whitespace after keyword
./omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py:415:5: E303 too many blank lines (4)
```

The PDF export has a few issues:

Screenshot 2024-11-29 at 10 45 11

Rdornier:
Hi @will-moore, 

Thanks for your feedback. I think I've corrected most of the bugs you reported.
There is only one that I wasn't able to reproduce. When I open/close/reopen the Edit ROI modal dialog, I didn't see the remaining label (I also tried to delete shapes, save and reopen).

Did you do someting particular to get this ? If you try again now, does it shows the same bug ?
Thanks.

will-moore:
Gave this another quick test. Most of the issues I had was clicking on a shape other than Rectangle which gave a `getText()` doesn't exist error.
Also I noticed that the Text is rendered with a Fill and Stroke which make it a bit BOLD and hard to read. Probably only need to have Fill for text and no Stroke.
In the ROI dialog, if I select a Rectangle that has some text, the Text field in the toolbar shows that text (as expected) but then if I select another Rectangle that has NO text, then I'd expect the Text input field to show no text (instead of still showing the previous text).

Rdornier:
Hi @will-moore , 

I've finalized the PR ; it's ready to be reviwed. 
All the points and bugs mentionned above should be done and fixed.

will-moore:
Here, I created the yellow and magenta shapes with labels, then I rotated the panel 90 degrees to the right. Then I Edited ROIs again and added the White arrow with label -> OK. Then I re-opened the Edit ROIs dialog to take this screenshot:

Screenshot 2025-04-08 at 16 13 08

The issue with the Arrow text being different orientation from the others did eventually resolve itself. 
There is a minor issue with the placing of the labels when panel is rotated, e.g. the magenta label is aligned to the TOP of the rectangle. So, it rotates around the middle of the label, which is 1/2 label width along the top side. But when the Rectangle is on it's side, the middle of the label is now 1/2 label width along the same side (which is now the Right side of the Rectangle). I think this is probably OK and would take too much effort to fix. Just wondering how to minimise the surprise factor for users.

Also, as you've mentioned before, there is the issue of different scaling of line widths (and now font-sizes) between the panel on the Figure and the panel in the ROI Editor. If you create a label in the ROI Editor that is e.g. 24pt so that it spans the panel, then when you `OK` and return to the Figure, it will be too big!
I wonder if we need to generate some scaling factor, based on the size of the panel at the point at which you open the Edit ROI dialog, that can be used to scale fonts and line-widths accordingly. E.g. if you open a very small panel, then a 18pt font in the ROI dialog is going to appear bigger on the image than if you'd opened a much bigger panel.
But this is just a thought and it would take some investigation... I'll think about it a bit...

will-moore:
I'm finding it a bit tricky to find nice Label placement for Arrows...
The outside edges are too far away (screenshot), and the inside corners overlap with the Arrow itself:

Screenshot 2025-04-08 at 16 57 44

Actually, there is a way that I can place a label where I want it: I create a Rectangle with "almost" invisible outline (0.25 thick), then I can drag this to where I want:

Screenshot 2025-04-08 at 17 10 42

I wonder if we should facilitate this "hack" by allowing users to create a "standalone" label? I think that users are really going to want to do that, or even have Labels with out any linked Shape.

But, to do that properly, we'd want a "Label" shape type in the ShapeEditor and for ShapeEditor to know where to put drag-handles etc.
I realise that this sounds like a fair bit more work, but I think it will actually be easier than trying to work out an "automatic" placement of labels that users find acceptable under all possible conditions.
Apologies for not suggesting it earlier, but I guess you only had labels on Rectangles to start, so this problem wasn't so obvious.

I think that you'd need a "Text" `T` toolbar button and in that mode, a single click would create a Label - maybe it's just a point with a line-width of `0`, so that when it's not selected you don't see it. This actually works quite well with the label positions allowing you to "anchor" the text to the point.

Screenshot 2025-04-08 at 17 17 07

The main issue is how to you click to select it if you can't see it?! Actually, that's probably something that needs fixing even for regular shapes and their labels. If I've got a label that mostly fills a shape, I can't select the shape by clicking on the label and I actually have to find a bit of the underlying shape to click. 

Screenshot 2025-04-08 at 17 43 54

One other point - it might be nice to have a label position of "Centre", which is useful for placing a label in the centre of an Ellipse or Rectangle.

Thanks for all the work; it's looking good. Let me know your thoughts and we can work out the best way of implementing this. Cheers

Rdornier:
Hi @will-moore 

Thanks for reviewing

> I'm finding it a bit tricky to find nice Label placement for Arrows...

Ok I see your point... The reason why I forced pre-defined positions was to stick with the behavior of labels around panels.

> I wonder if we should facilitate this "hack" by allowing users to create a "standalone" label? I think that users are really going to want to do that, or even have Labels with out any linked Shape.

I have some questions regarding this statement
1. is the 'standalone' label a floating label (i.e. a label that we can move as any shape) ?
2. If we don't link the label to any shape, how can we fix the positionning, as labels might be moved by hand ?

In order to address the different points (standalone, selectable and movable label), I thought about 2 solutions

#### Floating labels linked to existing shapes
Keeping most of what has been done so far, but introducing the selectable and movable options to each label
- Labels will be linked to a shape
- Each shape has only one shape linked (this is maybe too restrictive and we should maybe allow more than one label per shape. What do you think about that ?)
- Labels will be selectable (with handles, as any other shapes)
- Labels will not be forced on a pre-defined position => user can move it everywhere BUT some pre-defined positions are still available (including a new 'center' position)
- When the shape is deleted, the linked label(s) are also deleted
- Creating a label will be as simple as clicking on the "Text" T toolbar button. A default text is added to the shape at a default position but as it is now possible to move the label by hand, this should not be an issue. The text button should be disabled if no shapes is selected.
- To edit a label, one can select it and modify the text from the current textbar.
- Each label could now be deleted independentely
- If the label is moved, the shape is not moved
- If the shape is moved, all the labels are moved accordingly

The drawback of this solution is that we cannot create an indenpendant label, not linked to shapes. 

#### Floating independant labels 
In that scenario, the labels are completely independant from any shapes. All the attributes previously cited are kept, with the following updates
- Labels are not linked to any shapes
- Creating a label can be done any time, independently from if a shape is selected or not.
- Predefined positions will not be available anymore, as we cannot know which reference shape to take
- Labels and shapes can be moved independently from each other

The drawback of this solution is that it is not possible to stick with pre-defined positions (top, center, etc), as the labels are not linked a reference shape.

Another point is the support of TextROI imported from OMERO. In both solutions, it could now be supported.
- In solution n°1, it will be a label NOT linked to any shape, which is a bit strange regarding the rest of the labels, but possible
- In solution n°2, it will be a label as any others.

As long as I'm thinking about it now, maybe I can foresee a solution for combining both solutions into a single one. We could make a distinction between the two cases:
- One or more shapes are selected : in that case, the label is linked to the selected shapes and we fall into solution n°1. The icon of the text shape could be a `T`, with a drawing next to it.
- No shapes are selected : in that case, a label is created but not linked to any shapes and we fall into solution n°2. The icon of the text shape could be a single `T`.

image

Doing like this, I think it could properly address the label positionning.

> The main issue is how to you click to select it if you can't see it?! Actually, that's probably something that needs fixing even for regular shapes and their labels. If I've got a label that mostly fills a shape, I can't select the shape by clicking on the label and I actually have to find a bit of the underlying shape to click.

Yes, if I make the label selectable, this should be fixed

> maybe it's just a point with a line-width of 0, so that when it's not selected you don't see it. 

Maybe we won't need it anymore if the label can be selected and moved, right ?

> The issue with the Arrow text being different orientation from the others did eventually resolve itself.

Hum.. I'll have a look to this.

> I wonder if we need to generate some scaling factor, based on the size of the panel at the point at which you open the Edit ROI dialog, that can be used to scale fonts and line-widths accordingly. E.g. if you open a very small panel, then a 18pt font in the ROI dialog is going to appear bigger on the image than if you'd opened a much bigger panel.

I think it's a good idea.

Let me know what you think about, if it's good or not to proceed in that direction.
Thanks, and have a nice day.

will-moore:
@Rdornier I think we want "Floating independent labels" as well as "Shape with label" (as you have now).
This corresponds to the Shapes model in OMERO. Shapes can have text value and there is also a "Label" type of shape.

With a floating independent "Label", we need X and Y coordinates so you can drag it around. It could be nice to choose the "anchor" position for the label (where the label is positioned relative to the X, Y, coordinates). E.g. if the anchor is "left", then you can change the length of text in the label and the left-most edge of the text will stay in the same place. This is possible now if you add Text to a Point, except that you choose "Position -> outside edge/Right" to get the anchor to be "left":

Screenshot 2025-04-10 at 14 43 38

So, it may too confusing and complex so that we don't want to give users the option of "position" for standalone label. We just decide what the anchor behaviour is (e.g. left or centre) and then if the user needs to re-position it after editing the text then that's not a big deal.

I think a label is only created if you select the `T` icon and then click a point on the image (just the same as creating any other Shape). Maybe this should immediately give the new Label a text value of "Label" so that you can see it on the image and then edit the "Label" string to whatever you want.

Insight gives a default value of `TEXT` when you create a new label, and the anchor is the top-left of the label:

Screenshot 2025-04-10 at 14 55 06

Rdornier:
Hi @will-moore 

I've worked a bit on this and implemented what we discussed above.
I have an issue with rotated standalone labels being drawn on the PDF. There is a mismatch in position between OMERO and PDF but I couldn't find the reason of such difference.

image

I hope the rest of the implementation answers the different points.
Let me know what you are thinking about.

Rémy.

Tom-TBT:
This is impressive the amount of work you put into this. We need to be thorough with the testing, but this will make A LOT of people happy. I did tests of the current version.

* Free text label pre-set positions. For position in the other corner than `top left`, the label "disappear"
  * labels on the corners should be appearing in the respective corner
  * labels on the outer edge is logic that they are not appearing, but making it "disappear" is bad. To fix it you need to Select all, then position top left corner, then you get them back. But if you had done positioning work already for multiple labels, this would destroy it all.
  * Either the outer edge option are out for free text label, either it position them inside the panel.

* The button to update the text is called "Add". But it only makes sense to Add labels on shapes. In any other case, it's about "updating" the label.
  * Without shape selected, "Add" doesn't add free text label. It seems that this menu is useful only when a shape is selected (free text shape or any other shape to add label).
  * `Add` could be changed to `Show/Hide`. This would work also to have it prefilled with shape's `textValue` and show only on request (multiple selection would not apply text in text box, but just toggle the display). See next point.
  
* Adding ROI from existing shapes, default text should be the shape's `textValue`
  * If my ROIs have complicated label/name, I would like to have an automatic way to display the names in figure.

* Text color deserve it's own colorbox?
  * Label color is changed independently of it's shape, but opposite is not true: changing shape color changes the label color.
  * If I change a batch of shape's color, I then need to set back one by one the shape's text color. It's the only annoying use case I found.
  * Color menu for the text only would fit if the text form is placed below the image
  * I mean something like that: image

* Fill and fill opacity of independant text label should show a filled background behind the text. 
  * there is no easy way to add background to text (if we want it readable on EM images for example.
  * you would have to make a rectangle, add a label to it, and center the label manually

* The text of an ROI still shows when I deselect the ROI. It is confusing that after I edit the text and press enter, nothing happens (the shape is not selected)
  * Could the text label menu appear only when a shape is selected? 
 
  * The label edit form could be placed somewhere else, like below the image. The text functionalities are somewhat disjoint from the create/edit shape
  * Color menu for the text only would fit if the text form is placed below the image

* Z and T plane info on the top-left are disturbing
  * those should be put elsewhere, text labels are hard to select (or any other shape for the matter)
  * image


Again, the piece of work is amazing. It works really well already and I could do anything I want, but I think it would still benefit from these "tweeks".

Rdornier:
Thank you very very much @Tom-TBT  for your detailled and precise review and testing !
I'll make sure to address all of the bugs / tweaks you've listed

Tom-TBT:
Hi @Rdornier,
I gave a quick look, you improved it a lot. UI is intuitive and coherent. 💯 

One thing I noticed is that copy pasting a shape in the editor copies everything (shape and text fontsize, background color, ...) but not the text. 
Is it by design? Because the properties are preserved, it's not annoying. But I'm curious why text shouldn't be copied too.


I also have an error when loading polygons:
```
roi_loader_view.js:86 Uncaught ReferenceError: points is not defined
    at child.getParentShapeCoords (roi_loader_view.js:86:26)
    at child.addOmeroShape (roi_loader_view.js:58:46)
```

I don't see this error on my local master branch. Any idea what causes it?

Rdornier:
Thanks @Tom-TBT for your feedback !
I corrected the error on copy-pasting polygons.

Regarding the text which is not copied, this was at the beginning indended because of the design of the code (temp shapes are created, then deleted, and the destroy mechanism looked at which text is attached to shapes to destroy it as well).

But, as you mentionned, everything is copied execpt text. So, I found a solution to make a copy of everything. Now, when you copy-paste a shape, with text, the text is also copied.

Rémy.

will-moore:
This is looking pretty good, except that with PDF export I didn't see any labels on the figure.
Labels on the TIFF export are fine.

Tiny issue: When you `Hide` the label on a shape (button now says `Show`), then you select a different shape then re-select the shape with hidden label, the `Show/Hide` button says `Hide` instead of `Show`, so you have to click it twice to show.

Tom-TBT:
> This is looking pretty good, except that with PDF export I didn't see any labels on the figure.
Labels on the TIFF export are fine.

In my first attempt, I had a missing label on PDF and TIFF. Trying again after changing the labels, I could get all labels correctly. Not sure if I did a mistake with my first test, but since Will also mentioned missing label, that could be a bug. 
But I would need to figure out how to reproduce it.

If there's a bug, I think that would be in the JS/JSON, not in the python script (like a boolean set to False or a missing key).

Rdornier:
Hi @Tom-TBT , @will-moore 

Thanks for your feedback.
I wasn't able to reproduce the bug of missing labels on PDF. However, I fixed the tiny issue with show/hide, which may also affect the PDF.
If you try again, do you still have the PDF issue ? In case, can you describe the steps you've done to get it ?

Thanks,
Rémy.

Rdornier:
Hi @will-moore 

Thanks for the review.

> However, the bigger problem which we've discussed before (in relation to arrow head I think) is that the font sizes look very different in the ROI editor compared with on the page, depending on how big the panel is on the page.

Indeed, we've discussed about it in a separate issue #617 but I think it's good to correct the text scaling here. Thanks for the fix.

> One minor bug - having added some text label to an Ellipse, I then tried to remove the text, but if I deleted the text entirely from the input box so the input was "", then the update didn't apply.

I couldn't reproduce this small bug. When I delete the entire text from the input box, it is acutally updated.
I don't know if it is coming from my setup or not. Maybe if @Tom-TBT has time to test this, that would help me digging into that.

text-removal

BTW, I've fixed another small bug by not writing text in pdf if text is empty. Otherwise, if text had background, only background was printed in the pdf.

will-moore:
When I copy and paste a Shape with label, or just a Label, the font-size isn't scaled accordingly (this is fixed if I close and re-open the ROI dialog):

Screenshot 2025-09-24 at 12 38 26

I don't know if this is intended, but the standalone label has a "position" option (used for other shapes to position the text in relation to the shape). I see that it simply positions the label within the viewport. This was slightly unexpected, but I guess I can see a use-case for it.

Rdornier:
> When I copy and paste a Shape with label, or just a Label, the font-size isn't scaled accordingly (this is fixed if I close and re-open the ROI dialog):

Thanks for the review. I've fixed it.

> I don't know if this is intended, but the standalone label has a "position" option (used for other shapes to position the text in relation to the shape). I see that it simply positions the label within the viewport. This was slightly unexpected, but I guess I can see a use-case for it.

Yes, it was in the [wish list](https://github.com/ome/omero-figure/pull/600#issuecomment-2871839613) of @Tom-TBT and I also think it's a good idea to have a pre-set position for free text as well.

will-moore:
The selection of Shape-and-Label when you click the Label is working. Minor point: the Shape-edit controls (for Line-colour, Fill-colour, opacity, line-thickness) aren't updated to the selected Shape if you select the Shape by clicking the Label.
Here I click the Red-point (controls are updated) then I select the Cyan Shape by clicking the label - controls are out of sync:

Screenshot 2025-10-23 at 22 46 17

For the Undo bug - are you using `model.save(...)` instead of `model.set(...)` when updating flip/rotation etc? I think it needs `save()` to get added to the Undo/Redo queue.

Rdornier:
> The selection of Shape-and-Label when you click the Label is working. Minor point: the Shape-edit controls (for Line-colour, Fill-colour, opacity, line-thickness) aren't updated to the selected Shape if you select the Shape by clicking the Label.
Here I click the Red-point (controls are updated) then I select the Cyan Shape by clicking the label - controls are out of sync:

Corrected, thanks for pointing this out !

> For the Undo bug - are you using model.save(...) instead of model.set(...) when updating flip/rotation etc? I think it needs save() to get added to the Undo/Redo queue.

It looks like this is what I've done (see [those lines](https://github.com/ome/omero-figure/pull/600/files#diff-cd67fb35b25f63352d0aa138700d4eca5d96d03c8c8ebb6f8476ce5ab024ba09R1415-R1418)), unless I'm not looking in the right place

src/js/shape_editor/shape_manager.js
```diff
@@ -585,26 +785,53 @@ ShapeManager.prototype.deleteShapesByIds = function deleteShapesByIds(
...
       notSelected.push(s);
     }
   });
+
+  // intermediateShapes.forEach(function (s) {
+  //   if (shapeToTextDestroy.indexOf(s._id) > -1) {
+  //     s.destroyTextShape();
+  //   }
+  //   notSelected.push(s);
+  // });
```
will-moore: Could cleanup these commented-out lines, and those above

src/js/views/right_panel_view.js
```diff
@@ -1326,18 +1387,29 @@
...
+                    let newShapes = shapes.map(sh => {
+                        if (sh.type === "Text") {
+                            return {...sh, vFlip: isVerticalFlipped ? -1 : 1};
+                         }
+                        return sh;
+                    });
+                    m.save({'vertical_flip': isVerticalFlipped, 'shapes': newShapes});
+                }else{
+                    m.save('vertical_flip', isVerticalFlipped);
+                }
```
will-moore: This block can be reverted since the `panel_view.js` now updates the ShapeManager directly on render_shapes (as for isHorizontalFlipped below)

will-moore:
@Rdornier apologies I didn't spot those 4 issues in my branch - All just code cleanup really. Do you want to make those changes in another PR just now?

will-moore:
Opened https://github.com/Rdornier/omero-figure/pull/4

PR #602, opened 2024-11-28
Changelog 7.2.0

will-moore:
Only #592 is yet to be merged

PR #603, opened 2024-11-28
Inset bugfixes

### LINKED ISSUE #601: Bugs with Inset workflow
- When I copy and paste panels with insets, we assign new inset IDs so that the new panel + inset are still linked via new IDs.
However, if I do **another** copy and paste of the same panel + inset, I don't want these new panels to have the same new IDs as the others.

 - If I copy and paste an Inset panel (e.g. to make split-view panels of the inset) but then I delete one of them, the Inset ROI gets removed from the parent image, even though we still have an Inset panel 'linked' to it.

will-moore:
Fixes #601 

Test bug-fix 1:

 - Select a panel and use the "Inset" button to create an Inset
 - Copy and paste the newly created Inset panel to duplicate it
 - Now delete ONE of the 2 inset panels. The inset ROI shouldn't be deleted.
 - If you delete the other inset panel the inset ROI should be deleted as before.

Test bug-fix 2:

  - Select a panel and use the "Inset" button to create an Inset
  - Select both the original panel and the new inset panel.
  - Copy and Paste to duplicate the pair of panels
  - Repeat the copy and paste to duplicate the 2 panels again.
  - Check that each inset ROI is only "linked" to it's corresponding inset panel (that was duplicated at the same time).
 
Screenshot 2024-11-28 at 21 59 52

### DIFF
--- a/src/js/models/panel_model.js
+++ b/src/js/models/panel_model.js
@@ -102,7 +102,12 @@
                     // delete or update the "inset" Rectangle
                     let updated = this.get('shapes');
                     if (panelDeleted) {
-                        updated = updated.filter(shape => shape.id != insetRoiId);
+                        // Delete the inset Rectangle IF there are NO other remaining insets
+                        let insets = figureModel.panels.filter(p => p.get('insetRoiId') == insetRoiId);
+                        if (insets.length == 0) {
+                            updated = updated.filter(shape => shape.id != insetRoiId);
+                        }
+                        this.save('shapes', updated);
                     } else {
                         let rect = panel.getViewportAsRect();
                         updated = updated.map(shape => {
@@ -111,8 +116,8 @@
                             }
                             return shape;
                         });
+                        this.save('shapes', updated);
                     }
-                    this.save('shapes', updated);
                     this.silenceTriggers = false;
                 }
             }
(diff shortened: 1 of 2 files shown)


will-moore:
@Rdornier I wonder if you could have a look at these 2 bug fixes and 1 possible improvement? (see details above). Thanks!

I think that once this and #592 is merged, we should get the release out since we have a good amount of new stuff now (see changelog at #602).
I'll have a look at your ROI labels ASAP, and can do another release with that when it's ready. Thx

Rdornier:
@will-moore Sure !
- Bug 1 : OK
- Improvement : OK
- Bug 2 : OK to some extend.  If I do `copy > paste` and then `copy > paste`, then it is working properly. But If do `copy > paste > paste`, then the ROIs of the two copied panels are linked together  to the last copied ROIs (gray lines). If I move the central panel, nothing append on the central ROI


image

> I think that once this and https://github.com/ome/omero-figure/pull/592 is merged, we should get the release out since we have a good amount of new stuff now (see changelog at https://github.com/ome/omero-figure/pull/602).

Yes, I think it is good idea

> I'll have a look at your ROI labels ASAP, and can do another release with that when it's ready

Sounds good 👍

will-moore:
@Rdornier Thanks for catching that! Strange behaviour due to the clipboard data being a shallow copy for new panels. I think it's fixed with that commit above.

Rdornier:
Yes, it's fixed. 👍

PR #605, opened 2024-12-09
Colorbar feature

Tom-TBT:
This PR implements the optional display of a colorbar next to a panel. The colorbar shows the mapping between pixel values and LUT ("normal" color included). This is work from @MinaEnayat and me.
There are still adjustments to be made, but this is functional and mature enough that we thought it was ready to receive comments/ideas.

image

Implemented features:
* colorbar display showing the LUT of the first active channel (opinion on what other behavior when more than 1 channel?)
* The colorbar can be placed on one of the four sides of the panel
* Parameters for:
  * thickness of the colorbar
  * spacing between colorbar and panel
  * number of ticks+labels
  * ticks+labels+spine color
  * tick length
  * label margin (distance between label and dash)
  * label font size 
* Crop page around panels handles horizontal labels (measure the CSS property of generated labels). Could be adapted for the handling of right/left normal labels
* Export to pdf / tiff

A spine with the ticks and labels gives a visual clue when the LUT is the same color as the page background, but only on one side to avoid confusion with the LUT when the LUT is the same color as the spine:
image

Suggestions for tests:
* Intensity range:
  * 0-255 (full range)
  * 40-190 (different start-end value)
  * 16-bit range
* reversed channel flips the colorbar (not the labels)
* uneven panel resizing (width/height without keeping the ratio).
* Parameter testing thickness, spacing, label font, label margin, and tick length:
  * python export handles them all
  * "crop page around panel" handles them all
* all tests above for the python export to tiff & pdf

Known issues / work left:

- [ ] CSS display of ticks's dashes is uneven (not the case with the exported figure)
- [ ] Labels in python left/right justified, when JS has them centered (options?, possibility to rotate labels?)
- [ ] Labels in python do not always match the labels in JS (rounding difference)
- [ ] Handle left/right normal labels for "crop page around panels"
- [ ] Rearrange Colorbar form (adding/removing options ?)
- [ ] if color start-end range is too low, labels are duplicated from rounding (what about float range?)


src/js/views/panel_view.js
```diff
@@ -357,6 +365,56 @@
...
+            for (const chann of this.model.get('channels')) {
+                if(chann.active) {
+                    color = chann.color;
+                    start = chann.window?.start;
+                    end = chann.window?.end;
+                    reverseIntensity = chann.reverseIntensity;
+                    break;
+                }
+            }
+            var lut_url = await FigureLutPicker.loadLuts(false)  // Ensure lut url and list are loaded
```
will-moore, line 387: Wow, this looks great. Haven't tested yet, but just noticed you don't need the `false` here (or below) since that's the default.

Tom-TBT: I am not passing the parameter in the two occurences. I keet the parameter in `loadLuts()` in case, but it is not currently in use.

src/js/views/lutpicker.js
```diff
@@ -95,24 +125,8 @@ var LutPickerView = Backbone.View.extend({
...
-                this.lutsPngUrl = `${WEBGATEWAYINDEX}luts_png/`;
-            } else {
-                this.lutsPngUrl = STATIC_DIR + lutsPng;
-            }
-            this.lut_names = this.lut_names.map(lut_name => lut_name.split('/').pop());
-
-            this.render();
-        });
+        this.loadLuts();
+        this.render();
```
will-moore, line 129: This change now calls `this.render()` before the luts have loaded. Haven't tested to see if this actually causes a bug, but I think this will be fixed with:
```
this.loadLuts().then(() => this.render());
```

will-moore:
I wonder if you want to allow users to choose the channel for the colorbar (if you have more than one active)?
In this case, I'm most interested in the Green channel (but I still want the Red channel to be active):

Screenshot 2025-01-10 at 14 31 28

It's even possible to imagine someone wanting 2 or more colorbars at once, for different channels, but I guess that's a bit edge case (you would just do a split-view figure).
Actually, you can fake it by duplicating the panel and overlaying both panels, with the underneath panel having just one channel active! :)

Screenshot 2025-01-10 at 14 36 31

src/js/views/lutpicker.js
Tom-TBT: That may indeed, I added your fix. Thanks

Tom-TBT:
> I wonder if you want to allow users to choose the channel for the colorbar (if you have more than one active)?

> It's even possible to imagine someone wanting 2 or more colorbars at once

I assumed that colorbars only made sense with a single channel (especially if the channel uses a LUT). 
With your example, red and green overlapping produce a new color (orange) in neither of the two colorbars.

I guess your trick is a good enough workaround if someone has the use case, but I wouldn't make it the normal "use case" to show one colorbar per active channel.

Instead of adding a parameter for the channel to show a colorbar for, I would add a doc/tooltip somewhere indicating that the first active channel is used for the colorbar.

will-moore:
I had to try quite hard to "break" this, but I managed to paste some invalid text into the "Spacing" field, which gave me figure JSON `"spacing":null` and kinda broke the layout as the spacing is used to layout a bunch of stuff. This is really not too important to fix, as it's very edge case. But it did also expose that the `placeholder` attribute for the "Spacing" field is still `Height`.

Screenshot 2025-01-30 at 15 36 00

Actually, I'm not sure that any of the number field placeholder attributes are really useful as strings since they're mostly not readable and a placeholder should really be a suggestion of a default value? (I see that I've done it the same way with the Scalebar fields too, but I think this is also incorrect. See https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/placeholder).

The form layout needs tweaking a bit to get all the input fields to the same size and to align them to the left. I think you need `class="input-group-sm"` on all the divs that wrap the inputs, and break the form into 3 separate "row" elements (currently there are just 2 and the 2nd one wraps).

As discussed last week, I think we are running out of space in the `Labels` right-hand tab, so that any Labels are often hidden and you need to scroll to see them. I'm wondering if we should add a 4th tab to the right panel `Info, Preview, Labels, ???` but I'm not sure what to call it or what to put there (maybe Scalebar and Colorbar could go there). Maybe Inset and Border (and ROIs ???) can go too and just leave "Labels" in the Labels tab!?
Ideas welcome!

will-moore:
The LUT png is resized big enough to provide for bars of 50px thick, but if you go higher you get other ones showing. E.g. 100 wide:

Screenshot 2025-01-30 at 16 16 50

You could simply stretch the lut png a bit more, or limit the thickness. Probably both. NB: even with a thickness of 50 you get the next LUT creeping in as there's anti-aliasing at the join.
For most of the other parameters I think it's OK to allow users to choose crazy numbers if they want. With 1 "Tick" you get a `NaN` so maybe 2 is the minimum? Negative numbers don't make sense for most inputs, so maybe use `min="0"` or a higher min value if appropriate.

will-moore:
The PDF and TIFF export look great! Just maybe needs a thinker border line on the TIFF colorbar? Screenshot shows TIFF, PNG and figure:

Screenshot 2025-01-30 at 16 43 49

will-moore:
A couple more thoughts...

I tried batch editing 2 panels at once. In the previous screenshot, I selected the top 2 panels (blue and green) and wanted to make them consistent by choosing the same number of Ticks and Font size but not changing other properties such as position. 
However, when I make a change to any of the properties, *all* of the colorbar properties are applied to both panels. I was kinda hoping that I could make changes to each property individually, as you can with channels.

Then I checked the behaviour for Scalebar and found that this doesn't work Scalebar - If you make a change to one property, it tries to apply ALL of them (same as Colorbar). In some ways, Scalebar is worse because for values that are not the same between selected panels E.g. Length - it shows NO value when those panels are selected together. The idea here is that the UI lets you know when these values are different between selected panels (which Colorbar doesn't do). BUT, when you edit a different property (E.g. Height) the `Length:null` gets saved to both panels!

So, it's probably best just to leave it as is. In the scenario above, I was actually happy to have all the properties consistent between the 2 colorbars, *except* for position. So I just had to re-select the first panel and reset it's colorbar position, which wasn't a big deal.

Tom-TBT:
1.  I removed the placeholders of the input fields

2. I stretched the png to 1000px, but that caused artifacts because of the anti-aliasing (seen with lower values too). → I added the CSS property `image-rendering: pixelated;` everywhere the LUT png is used. Here's the before/after (when the colorbar is wider than 1000px):
image
image

The artifact I mentioned for the LUT "Edges" (normal on top)
image
image
The yellowish thing is unrelated (it's in the LUT from the server, maybe an issue)


3. Fixed the issue with tick length layout on export

4. Fixed the display issue on export for multi-page

5. Issue with line thickness in tiff came from this (works for horizontal lines only):
https://github.com/ome/omero-figure/blob/9cc36cde05bde4def7b62b07c2e9ae5f66712e96/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py#L2457-L2458
Fixed by adding the same logic for vertical lines

6. Refactoring of the colorbar form, work in progress: 
image
* I rearranged the parameters and renamed some (only in the form until it's settled):
  * Spacing → Pad
  * Length → removed the parameter name on the form
  * Label margin → Margin
The first row affects the size and location of the colorbar, the second row is to control the ticks. 
Space is scarce hence why I removed the "length". Instead I could remove the "margin" parameter as I don't think it's useful (fine with the value `0` IMO).

will-moore:
All looks great, thanks.
Maybe "Gap" instead of pad? (we also use "Gap" for the grid spacing between panels).

I think the form layout is nice. A couple of the inputs could use a tool-tip "title" to help clarity. E.g. "Number of ticks" and "Tick Length".

For me, the Length and Margin fields are a tiny bit small if you have a larger number (20 etc) in them:

Screenshot 2025-02-14 at 11 06 08

I know that space is tight but I found that going from a width of `33px` to `35px` for those fields was enough to fix that effect and not be too long for the row.

Otherwise everything looks good.

Tom-TBT:
Thanks Will. 
my todo status now:
* rename the variable to the GUI names
* add tooltips
* test with different layout with a fourth tab
* There's an offset with the ticks in the tiff export of 0, 1 or 2 pixels. Some rounding issue but I was not able to figure it out... My understanding as of now:
  * `def scale_to_export_dpi(pixels):` floors the values instead of rounding
  * `paste_image` adds scaled `x` and `width` whereas `draw_scalebar_line` scales `x` and `x2` (`x2 = x+width`)
  * scaling x and x2 makes more sense to avoid a double-rounding (`round(x2)` is closer to the true position than `round(x) + round(width)`)

omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
```diff
@@ -2454,8 +2681,8 @@ def draw_scalebar_line(self, x, y, x2, y2, width, rgb):
         y2 = scale_to_export_dpi(y2)
         width = scale_to_export_dpi(width)
 
-        for value in range(-width // 2, width // 2):
-            draw.line([(x, y + value), (x2, y2 + value)], fill=rgb)
+        # the coordinates are included in the line
+        draw.line([(x, y), (x2, y2)], fill=rgb, width=width)
```
Tom-TBT, line 2686: That change was first to try to fix that pixel offset, but I realized it could be useful to not have independent lines object in the PDF. 
Note however that result would be different if x and x2 or y and y2 are swapped.
I will add code to ensure x2 < x and y2 < y (i think that's how it should be reproduce previous behavior)

omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
```diff
@@ -2558,6 +2889,11 @@ def draw_text(self, text, x, y, fontsize, rgb, align="center"):
             x = x - (temp_label.size[0] / 2)
         elif align == "right":
             x = x - temp_label.size[0]
+
+        if align not in ["left-vertical", "right-vertical"]:
+            # The text in TIFF is higher compared to PDF. Add offset
+            y = y + scale_to_export_dpi(1)
+
         x = int(round(x))
```
Tom-TBT, line 2897: added this because I saw an offset between text in pdf and text in tiff. Need to validate that it works for:
* labels
* scalebar
* colorbar

omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
```diff
@@ -1677,6 +1676,232 @@ def draw_scalebar(self, panel, region_width, page):
...
+        self.paste_image(pil_img, img_name, ramp_d, page, dpi,
+                         is_colorbar=True)
+
+        # Handle page offsets
+        ramp_d['x'] -= page['x']
+        ramp_d['y'] -= page['y']
+        labels_x = numpy.array(labels_x) - page['x']
+        labels_y = numpy.array(labels_y) - page['y']
+
+        self.draw_colorbar_ticks(colorbar, ramp_d, labels, labels_x, labels_y)
```
Tom-TBT, line 1829: that's where I had to split the code between PDF and TIFF export. With the TIFF, I need extra code to handle pixel offsets.

omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
```diff
@@ -2395,26 +2624,24 @@ def paste_image(self, pil_img, img_name, panel, page, dpi=None):
...
         y = scale_to_export_dpi(y)
-        width = scale_to_export_dpi(width)
-        height = scale_to_export_dpi(height)
-
-        x = int(round(x))
-        y = int(round(y))
-        width = int(round(width))
-        height = int(round(height))
+        width = x2 - x
+        height = y2 - y
```
Tom-TBT, line 2632: scaling the width/height independently can produce pixel offsets because of the rounding.
`round(x) + round(width) != round(x + width)`

But that created another problem that can change the ratio `width/height` (if `height == width` before rounding, this is not always true after the rounding because of the decimal values of x and y)
That's something I need to fix

will-moore (commented):
Sorry, forgot to "finish" my review yesterday

will-moore:
Thanks, that fixed the Figure export on my Docker server 👍

omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
Tom-TBT: PDF and TIFF look the same after all adjustments of vertical label positions.

omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
Tom-TBT: done

omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
Tom-TBT: In the end, I think it's better to leave it this way. Square panels would have a difference of maximum one pixel between width and height.
The only way I thought of handling it would require updating the position of the panels beforehand, then make sure alignments remain, ...   too much code for something barely noticeable IMO

Tom-TBT:
Hi Will, with these last changes I have everything I wanted.
So it's ready for whenever @pwalczysko or you have time to give it a final🤞 review.

will-moore:
I noticed that there's a lot of calls to `webgateway/luts` when I load a figure that has a bunch of colorbars.

Screenshot 2025-03-13 at 10 15 28

It's because every panel's `render_colorbar()` is calling `var lut_url = await FigureLutPicker.loadLuts()` at the same time, even though most of them aren't using it. This is cached after the first call, so it's only because we do an initial render of all the panels at once when we see this.

The nicest solution is to load the LUTs before we try to render the figure. Doing this in `LutPickerView.initialize()` seems to work pretty well. This doesn't *guarantee* that LUTs will be loaded before we try to render the panels, but it should help a lot:

```
+++ b/src/js/views/lutpicker.js
@@ -38,6 +38,7 @@ var LutPickerView = Backbone.View.extend({
     initialize:function () {
         this.lutModal = new bootstrap.Modal('#lutpickerModal');
         this.lut_names = [];
+        this.loadLuts();
     },
```

The other change that I tested as part of this was to only `loadLuts()` in the panel rendering if we need to:

```
+++ b/src/js/views/panel_view.js
@@ -384,9 +384,13 @@
                     break;
                 }
             }
-            var lut_url = await FigureLutPicker.loadLuts()  // Ensure lut url and list are loaded
-            var lutBgPos = FigureLutPicker.getLutBackgroundPosition(color);
+            var lut_url;
+            var lutBgPos;
             var isLUT = !(/^[0-9a-fA-F]+$/.test(color))  // check if it's a normal color or a LUT
+            if (isLUT) {
+                lut_url = await FigureLutPicker.loadLuts()  // Ensure lut url and list are loaded
+                lutBgPos = FigureLutPicker.getLutBackgroundPosition(color);
+            }
```

This will have no effect if the LUTs are already loaded (by the fix above), but it seems a trivial change that *could* avoid unnecessary calls if the LUTs haven't already loaded?

omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
```diff
@@ -1677,6 +1676,234 @@ def draw_scalebar(self, panel, region_width, page):
...
+                              * numpy.array([r, g, b], dtype=numpy.uint8).T)
+                color_ramp = color_ramp.astype(numpy.uint8)
+            except ValueError:
+                pass
+
+        else:
+            script_service = self.conn.getScriptService()
+            luts = script_service.getScriptsByMimetype("text/x-lut")
+            self.conn.SERVICE_OPTS.setOmeroGroup(  # required to get LUT files
+                self.conn.getEventContext().groupId)
```
will-moore, line 1706: Testing just now, I found that after creating a LUT colorbar, other images failed to load and were missing from the PDF.
So I realised that we need to `setOmeroGroup("-1")` after loading the originalFile, otherwise we might fail to load subsequent images.

omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
```diff
@@ -2683,6 +2683,9 @@ def draw_scalebar_line(self, x, y, x2, y2, width, rgb):
         y2 = scale_to_export_dpi(y2)
         width = scale_to_export_dpi(width)
 
+        x, x2 = (x2, x) if x > x2 else (x, x2)
+        y, y2 = (y2, y) if y > y2 else (y, 2)
```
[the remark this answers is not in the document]
Tom-TBT: 😬 sorry    fixed

omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
Tom-TBT: I could not reproduce, but fixed. Tell me if that worked

will-moore: Yes, that worked. You'll only see the bug if your images aren't in your default group.

will-moore (approved):
Thanks @Tom-TBT. Did another round of testing and everything is looking very nice.
Great feature to have - Thanks for all the work 👍

PR #607, opened 2025-01-08
fix NoneType exception on timeincrement

### LINKED ISSUE #606: Timelapse images cannot be added in any new figure
Hello

I recently noticed that specific images (timelapse images) could not be added in the omero.figure anymore. An error message like the one below is shown

Rdornier:
Fixes #606 

Added a `None` check in [that line](https://github.com/ome/omero-figure/blob/master/omero_figure/omeroutils.py#L76)

### DIFF
--- a/omero_figure/omeroutils.py
+++ b/omero_figure/omeroutils.py
@@ -73,7 +73,7 @@ def get_timestamps(conn, image):
         except Exception as error:
             print(f"An exception occured: {error}\n"
                   "maybe the image has no 'timeIncrement' set")
-        if converted_value != 0:
+        if converted_value != 0 and converted_value is not None:
             for i in range(image.getSizeT()):
                 timemap[i] = i*converted_value
 


will-moore:
Thanks for the report and fix.
I can't seem to reproduce this on any images I've tested with (I can't even set the timeIncrement value to None on a test image using example at https://forum.image.sc/t/how-to-correctly-set-a-timeincrement-for-omero-pixels/101600).
Do you see this on all time-lapse images?

However, the fix looks like the right fix for this error, so I'm happy to approve...

will-moore (approved):
Deployed on merge-ci and tested loading various time-lapse images. Images and Timestamps loading without errors.

will-moore:
Released as omero-figure 7.2.1

Rdornier:
Thanks a lot for the quick merge & release, Will !

> Do you see this on all time-lapse images?

No, only on specific images coming from an Incucyte microscope.

> I can't seem to reproduce this on any images I've tested with

I cannot share the problematic data. However, if you create a timelapse image on Fiji, you can get the same error.
Here is a dummy timelapse [timelapse-test.zip](https://github.com/user-attachments/files/18372578/timelapse-test.zip) that should produce the bug.

PR #610, opened 2025-01-22
Add docs/contributing.rst

will-moore:
Various notes that might be useful for contributors


will-moore:
cc @Tom-TBT @Rdornier 
Any useful tips or questions that could be useful to have in these notes?
E.g. development gotchas or things that took a while to work out?
I haven't included a lot of the basics as you seem to have figured it out yourselves, but if there's something that was confusing where docs could help, it would be good to cover. Thanks.

Tom-TBT:
Maybe adding development environment example? (thinking of weird people like me doing it in Windows).

* Vite running from the Ubuntu WSL
* OMERO.web running locally in Docker (and connecting to an outside server, but could be in Docker too)
* Coding in VS Code (on the same folder as where Vite runs)

Rdornier:
> Maybe adding development environment example? 

I agree, I spent a lot of time configuring my dev env with docker.
Here is my internal doc I've made to remember what to do, in case I had to redo it (I think it doesn't explain how and where to install Vite, but the rest should be detailled).
[omero-figure-doc.md](https://github.com/user-attachments/files/18521191/omero-figure-doc.md)

In the `code notes`, I think we could add that
- All files under `omero_figure/static` and `omero_figure/template` should not be modified, as they are created during the build.
- The `Figure_to_PDF.py` script is located under `omero_figure/scripts/omero/figure_scripts`. If modified, and to test changes, you can replace the script from the webclient (located under the menu `figure-script`) and update omero-figure.

On top of that, maybe we could also show an 'overview' of 'How works omero.figure ?', for example, with a map like this, to quickly see the interactions between user/code/file/shapes...
[ConceptMap_OMERO_Figure.pptx](https://github.com/user-attachments/files/18521341/ConceptMap_OMERO_Figure.pptx)

image


The rest looks pretty nice to me, and I like the `testing` section 👍

PR #623, opened 2025-02-12
Fix wrapping of label X button. see issues/622

### LINKED ISSUE #622: Labels panel - swap and delete icons
Figure is

https://outreach.openmicroscopy.org/figure/file/153178

To repeat:
1. Chrome on Mac
2. Make sure you are on 90% zoom in Chrome
3. Go to https://outreach.openmicroscopy.org/figure/file/153178
4. Select the two leftmost panels in first row (they have labels on them).
5. Select the Labels tab in the control panel.
6. Observe that the up-down arrow icons and the red-cross icons do not match the label lines.
7. Observe that this mismatch makes it impossible to delete some labels using the red-cross icon
8. Note: it is not that the red-cross icons are just on wrong lines, even with the best of efforts, you cannot delete the row 3 from the top (the cross below it does not work)

Edit: After some effort, I have managed to delete all the labels. I found out that the topmost red-cross is kind of a "master-delete" and if you keep pressing on it, the lines below keep vanishing one-by-one. Although the deletion is done now, this is not the desired behaviour -> maybe just take it as. a hint to find out what is happening.

Image

will-moore:
Fixes #622.

Test edit labels form with browser on 90% zoom as described on the issue #622.

### DIFF
--- a/src/css/figure.css
+++ b/src/css/figure.css
@@ -203,7 +203,6 @@
         padding: 15px;
     }
     .close {
-        float: right;
         font-size: 21px;
         font-weight: bold;
         line-height: 1;


pwalczysko:
Tested on FF and Chrome (the same figure exported from outreach) https://merge-ci.openmicroscopy.org/web/figure/file/16453 
The same zoom in Chrome (90%). 

lgtm

PR #638, opened 2025-04-25
Fix arrow opacity

### LINKED ISSUE #620: Arrow head FillOpacity in PDF export
The "Arrow" shape when stored in the Figure JSON doesn't include FillOpacity, but it seems that other shapes with FillOpacity can cause the Arrows to be exported in PDF with less than full FillOpacity:

If the Rectangle is deleted, the Arrow Head is restored to full opacity in the PDF.
I don't see this bug in TIFF export.

Image

Rdornier:
Fixes #620 

Left: OMERO
Right: PDF

image


will-moore (approved):
Works great, thanks

PR #641, opened 2025-08-13
OMERO ROIs improvements : filtering visible and adding all

### LINKED ISSUE #628: Loading all ROIs from OMERO instead of one-by-one
In the `edit ROI` panel, there is no easy way to import ALL ROIs at once in the figure. We can only select them one by one, which can be a long process if there are dozens of them.

Image

Maybe adding a button `Add all ROIs` could be a good idea.

Rémy.

Rdornier:
Fixes #628 

When clicking on `Load ROIs`, only ROIs that can be added (i.e. intersecting the viewport) are displayed in the list. This makes the interface more clear and when the mouse is passed over the ROIs, they can all be seen. The number of available ROIs vs total ROIs is also displayed.

Additionnally, a button `Add all` is added to insert all visible ROIs at once, to not have to add them all manually.

image
</br>

image


Rdornier:
Hi @pwalczysko 

Thanks for your feedback. I wsn't aware that this was already tested. I was testing with 10-20 ROIs and it worked fine. I've given a try with 280 and now, I begin to see a bit of lagging when adding them all (15s). Also, after saving them, when changing dimensions of the panel, the responsivity is also not the best.

One thing I could add it to enable `Add all` option only if the number of visible ROIs is lower than a certain number (ex: 200), to avoid having such lagging issue. And also working only on the current page (not loading all ROIs from all pages).

I really would like to add this feature to add all ROIs at once because this is a scenario that is asking more and more often. Users are using QuPath and would like to display annotations, of a with a certain name, in their figure. That's why I also would like to implement the text filter.

Rdornier:
> I think the nicest way to do the filtering is on the server, so that we don't get the strange pagination behaviour above (showing only 1 ROI from the first page).

I agree, yes. That would be nice. What about the filtering by ROI comments ? That's something I also would like implement because it can be helpful. I guess this filtering will also be part of the sql query, right ?

will-moore:
I'm afraid I'm going to close this for now; I don't think it's going to get in anytime soon, and we are about to start a big code move with renaming / moving the app: see https://github.com/ome/omero-figure/issues/689. 
We'll need to revisit this issue when that's all complete.

Rdornier:
Ok, I see. Big changes indeed...

PR #645, opened 2025-10-31
Fix word wrapping in #info-tab-pane

JensWendt:
I updated recently from 6.2.2 to 7.2.1 and immediatly I had users coming to me complaining about not being able to see their full (needlessly long -.-) image names. Apparently this changed in the last versions, maybe because of migration away from jquery.
Inspired by the PR https://github.com/ome/omero-web/pull/574 for OMERO.web I propose a similar implementation for Figure.

### DIFF
--- a/src/css/figure.css
+++ b/src/css/figure.css
@@ -1,5 +1,3 @@
-
-
 /* Copyright (C) 2014 University of Dundee & Open Microscopy Environment.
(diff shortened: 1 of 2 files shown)


JensWendt:
okay, I f*ed up with auto-format and git. I will just open another PR, because I am unable to recover my mess

PR #649, opened 2025-12-04
Replace manual changelog by automated GH release step

jburel:
[Description of PR]

Releases are usually not created when a new tag is pushed meaning that the status page reports all releases

Screenshot 2025-12-04 at 11 50 31
The workflow introduced in this PR is similar to what has already done in other repositories e.g. https://github.com/ome/omero-marshal/blob/master/.github/workflows/publish_pypi.yml#L23


jburel:
@will-moore I have now linked a release to every tag using the "Generate release notes" features in the UI
This is the strategy used in all the repositories that we are updating e.g. https://github.com/ome/napari-omero/releases/tag/v0.6.0

I don't think we need to keep the changelog but I understand if you wish to keep for a bit.

will-moore (approved):
Thanks for creating all the releases. Looks like we don't need CHANGELOG now.

jburel:
gfi

PR #651, opened 2025-12-04
More compliant align to grid

### LINKED ISSUE #621: Align-To-Grid algorithm is too conservative
Align to Grid starts at the top-left selected panel and tries to create rows, looking for the next panel to the right to align and if none is found then it moves to the next row, starting below the first panel in the previous Row.

If no panel is found then it stops.

This means that panels *before* the first column (of the top-left panel) get missed, and so do any that come after a "gap".

E.g. Before "Align To Grid"...

Image

and After...

Image

In general, the existing algorithm (align each panel to it's neighbour in rows, rather than one overall grid) works well for handling panels of different sizes.
We just need to look a bit harder to find the first panel of each row:

 - check the column to the left
 - for each panel of the previous row, check below until we find the first panel of the next row

Rdornier:
@will-moore 

Fix #621 

This 

image

Becomes that 

image

### DIFF
--- a/src/js/models/figure_model.js
+++ b/src/js/models/figure_model.js
@@ -1,6 +1,6 @@
 
     import Backbone from "backbone";
-    import _, { last } from 'underscore';
+    import _ from 'underscore';
     import $ from "jquery";
 
     import {PanelList, Panel} from "./panel_model";


will-moore:
This is really nice, thanks.
Just one small change I can suggest...

In the case where the user has tried to create a "sparse" grid:

Screenshot 2025-12-05 at 08 28 57

this is currently lost:

Screenshot 2025-12-05 at 08 29 05

whereas I think it would be nicer to preserve the layout a little better:

Screenshot 2025-12-05 at 08 33 38

This is the change that seems to work for me, if you think that makes sense and I'm not missing something? (I didn't fully look into the `lastRefColumn` logic there)

```
+++ b/src/js/models/figure_model.js
@@ -845,16 +845,15 @@
             // update position of panels 
             for (var [r, y] of Object.entries(rows_position)){
                 var row = grid[r];
-                var closest_column = this.get_closest_column(row[0], reference_grid, last_panel_width)
-                if(closest_column >= 0){
-                    new_x = reference_grid[closest_column]
-                }else{
-                    var lastRefColumn = Object.keys(reference_grid).length - 1
-                    new_x = reference_grid[lastRefColumn] + last_panel_width + spacer
-                }
-
                 for (var c=0; c<row.length; c++) {
                     let panel = row[c];
+                    var closest_column = this.get_closest_column(panel, reference_grid, last_panel_width);
+                    if(closest_column >= 0){
+                        new_x = reference_grid[closest_column]
+                    }else{
+                        var lastRefColumn = Object.keys(reference_grid).length - 1
+                        new_x = reference_grid[lastRefColumn] + last_panel_width + spacer
+                    }
                     panel.save({'x':new_x, 'y':y});
                     new_x = new_x + spacer + panel.get('width');
                 }
```

Rdornier:
Hi @will-moore 

Ok I see the point.
I modify a bit your changes, as it was only working if the row with the most left panel was full.
Should be ok now. You can give it a try and see if it do the job correctly.

Rémy.


image

image

src/js/models/figure_model.js
```diff
@@ -742,53 +742,171 @@
...
+                reference_grid = {};
+
+            // start aligning the row with the most left panel
+            // This row is considered as a reference to align
+            // the rest of the panels
+            var global_index = 0
+            for (var c = 0; c < ref_row.length; c++) {
+                let panel = ref_row[c];
+                var current_x = panel.get('x')
+                var gap = Math.floor(Math.abs(current_x - new_x) / last_panel_width)
```
will-moore: On the first iteration through this for loop, `last_panel_width` is `0`.
If `current_x` and `new_x` are equal then `gap` is `NaN` and the inner for-loop below doesn't get executed, but if `current_x` and `new_x` are not equal then `gap` is `Infinite` and the inner for loop below goes on "forever".

I came across this bug when I randomly tried to put 2 overlapping panels into a grid:

Screenshot 2025-12-18 at 15 44 15

will-moore: @Rdornier - you only see this bug if the left-most panel is lower than the other one.
I think we just need to skip a bunch of logic for the first iteration through the loop?

PR #654, opened 2025-12-09
Refresh metadata button to reload channel names, pixel sizes

Tom-TBT:
This PR introduces a button to refresh the metadata of all present panels (instead of a workaround consisting in replacing the image IDs with their own to trigger the refresh).

I hesitated between refreshing only selected panels versus all, but from a user experience I thought refreshing all may be less confusing ("I hit the refresh button but I can't see anything happening")

The loading only happens once per image ID and applies the fetched metadata to all corresponding panels.

Tests:

* Run $ npm build and test from your omero-web deployment instead of the vite.js dev server.
**built and behaves the same. All tests below on the omero-web deployment**

* Check that you can open old figures (created before your feature was added).
**can open and feature works**

* Add several panels to the figure, selecting each in turn and check the feature is working (for example, editing a particular property of each panel).
**works for more than 1 panel per image ID for multiple image IDs**

* Select several panels at once, each with different values of this property.
**feature apply to all panels regardless of current selection**

* Check it is the displayed correctly and batch-editing is enabled if appropriate.
**applies to all panels**

* Test that Undo and Redo work as expected.
**Undo Redo works for batch of panels (not one at a time). Undo would put back old metadata, but if confusion happens, refreshing gets to the latest metadata state.**

* Is the Save button enabled when you make a change?
**I had to make an explicit call to enable the save button:** `this.set('unsaved', true);`
**I don't know why updating the panels model isn't enabling it. Feedback needed**

* If you refresh the page after Saving, does the change persist?
**yes**

* When you export the figure to PDF and TIFF, is this feature handled correctly?
**yes**

* Make sure the export script can handle panels where this new property is missing: E.g. use panel.get('property', default).
**no new properties**

### DIFF
--- a/src/js/models/panel_model.js
+++ b/src/js/models/panel_model.js
@@ -613,8 +613,17 @@
                 }
             } else {
                 // project, dataset, screen, plate, well, field (name, id, label, index)
-                var parentVal = this.get("parents")?.[property]?.[format];
-                text = parentVal == null ? "undefined" : "" + parentVal;
+                if (this.get("parents") == null && property === "dataset"){
+                    // fallback on old property
+                    if (format === "id") {
+                        text = ""+this.get('datasetId');
+                    } else if (format === "name") {
+                        text = this.get('datasetName') ? this.get('datasetName') : "No/Many Datasets";
+                    }
+                } else {
+                    var parentVal = this.get("parents")?.[property]?.[format];
+                    text = parentVal == null ? "undefined" : "" + parentVal;
+                }
             }
             return text;
         },


will-moore:
Thanks - this is working with no bugs. A couple of points...

 - I think that the "real-estate" is a bit limited at the top toolbar for a big button (or even a small button). Maybe this could be moved under the `File` menu as it seems similar to other whole-figure actions there like `Paper Setup` and `Add Figure Legend`. I realise that the "discoverability" of the feature is not so great there, but I think interested users will still find it.
 - Maybe the tooltip could give slightly more info about what will be updated. E.g. "Reload metadata such as image names, pixel sizes and channel names from the server, for all panels"
 - It's a little hard to know when the process is complete. I think a full dialog is probably a bit of an overkill but even an `alert("Updated metadata for 10 images")` would be nice to know that it worked.

src/js/models/figure_model.js
```diff
@@ -621,6 +621,77 @@
...
+                            panel.set({
+                                'name': data.meta.imageName,
+                                'datasetName': data.meta.datasetName,
+                                'datasetId': data.meta.datasetId,
+                                'pixel_size_x': data.pixel_size.valueX,
+                                'pixel_size_y': data.pixel_size.valueY,
+                                'pixel_size_z': data.pixel_size.valueZ,
+                                'pixel_size_x_symbol': data.pixel_size.symbolX,
+                                'pixel_size_z_symbol': data.pixel_size.symbolZ,
+                                'pixel_size_x_unit': data.pixel_size.unitX,
```
will-moore: For images that don't have pixel size info, this is `undefined` which gets set on the model, overriding the default value of "MICROMETERS".

This causes an error with `LENGTH_UNITS[pixel_unit].microns` since `pixel_unit` is `undefined` at
http://github.com/ome/omero-figure/blob/88b01ef56d2140b716e7f7022a5e7bd3d4b2c654/src/js/views/panel_view.js#L154

This *doesn't* happen when the panel is originally created, probably because any `undefined` values passed into `self.panels.create(n..` are ignored and the default values for the model are used instead. Default for `pixel_size_x_unit` is `MICROMETERS`.

So, you can "show" a scalebar for an Image with no pixel sizes - you just get a scalebar with `NaN` length, but no JS error. But when you refresh metadata or setId for a similar image without pixel sizes, then you DO get an error.

NB: this same bug already exists in `setId()`, so it would be good to fix it in both places, thx.

Tom-TBT:
I changed the way pixel sizes are updated. I start by unset previous values, set the defaults MICROMETER in `newData`, and only if the pixel_size are defined, they are added to `newData`

Tom-TBT:
Thank you Will for these.
* `panel.trigger('change:labels', panel);` I couldn't reproduce myself so I'll trust you with that. It still works as I'd expect
* `if (data.pixel_size)`  yes that was really wrong. I reverted to old code, and only added logic to set unit and symbol if X pixel size is undefined.
* `[dataset.name]`  I now check if parents exists. If not, it will use the old properties datasetId and datasetName
* I also found missing from the refresh that X and Z pixel units where not set from the data object. So I added these two

PR #655, opened 2025-12-09
Label on rois

will-moore:
Work in progress... Investigating...

Currently ONLY the `ShapeEditor` code and the `shapeEditorTest` code has been updated, but not the main app code.

To test:

```
$ npm run start
```
Go to http://localhost:8080/shapeEditorTest.html

This builds on #600, aiming to simplify the association of Text labels with other shapes:

 - The only Shape that supports linking of a Text shape is Rectangle - This is specified simply by the `labelId` attribute (as before in #600)
 - All other Shapes, including `Text` shape itself have no references to other shapes
 - All `Text` Shapes are positioned purely by `x` and `y` coordinates. They may have `textAchor` (e.g. `start`, `middle`) to position the text relative to the `x,y` coordinate. The idea is to simplify the model and also make it easier to reproduce for PDF and TIFF export
 - When the Rectangle is selected, it also selects the `textShape` if any exists (same as before)
 - The `textShape` can be selected on it's own and dragged to reposition, relative to the Rectangle, as desired
 - It's position relative to the `top-left` of the Rectangle is maintained when the Rectangle is resized (don't allow users to set any other "positions")
 - The `textShape` position is maintained relative to the centre of the Rectangle when the Rectangle is rotated (NB: this is currently handled by the `shapeEditorTest.js` code, not the `ShapeManager` or Rectangle code.
- The `strokeColor` sets the text Color and the `fillColor` sets the background color of the Text.
- I also added various controls to the `shapeEditorTest` page so we can test more things there

To test `shapeEditorTest.html`...

 - The current 2 labels have their `x,y` coordinates set to the same as the Arrow points. One label has `start` anchor and the other has `middle`.
 - If you rotate the image with the slider, you can see that the labels rotate about the arrow points.
 - Click "Add Inset" to create Inset(s).
 - Click an Inset Rectangle - the Label will also be selected. Can then drag them together (default behaviour for multiple selected shapes), update the strokeColor of both etc. Can also set the Text label & font size (will apply to all the currently selected labels)
 - Click on an Inset label to select just the label. Can drag to reposition with respect to the Rectangle (e.g. to a different corner or to tweak position for a bigger font size etc) or set a different colour, background/fill colour etc.
 - Try to rotate the Inset (slider) and the label should rotate in sync around the same centre-point. The `A` label uses `middle` text-anchor so it should rotate around the middle of the letter.
 - If you Resize the Inset Rectangle, you'll notice that the label maintains it's position with respect to the top-left corner
 - Create new Text Shape(s) and edit them

Questions

 - Do we need to let users choose the `text-anchor` for non-inset labels? It only has an effect when you resize or rotate the panel or change the Font-size. We could start by initially supporting only `start` and see if that feels limiting?
 - I think that only linking the Inset label to the top-left of the Rectangle is OK, since that should be OK for most users and if you do want to move it to a different corner you still can drag it to where you want - You just might have to tweak it yourself if you resize the panel. Sound OK?
 - Any objections to me going ahead and start updating the rest of the #600 code to adopt this behaviour?

cc @Tom-TBT @Rdornier

### DIFF
--- a/src/js/views/roi_modal_view.js
+++ b/src/js/views/roi_modal_view.js
@@ -450,7 +450,7 @@ export const RoiModalView = Backbone.View.extend({
             }
             this.shapeManager.setState(newState);
             if (this.shapeManager.getSelectedShapes().length == 0){
-                this.shapeManager.setText("");
+                this.shapeManager.setText("Text");
             }
             this.renderToolbar();
         },


will-moore:
More testing (before the padding change) - the wonky `B` in the first screenshot is fixed by a commit above

Screenshot 2025-12-12 at 15 33 25

Screenshot 2025-12-12 at 15 30 50

will-moore:
Closing this in favour of https://github.com/Rdornier/omero-figure/pull/3 to get these changes back into #600

PR #656, opened 2025-12-18
Option to add roi comment when adding ROI from OMERO

Tom-TBT:
I added a checkbox to give the possibility to load the shape comment when an ROI is added from OMERO:
image

The checkbox is currently checked by default, and the properties of the text are assigned by default to the values of the ShapeManager (`fillOpacity`, `fillColor`, `fontSize`, `fontColor`, `textAnchor`)

The iviewer places the comments differently for polylines and polygon, so I think I should change the default XY position.
How the iviewer shows the shapes:
image

The comments in iviewer also have a fontSize, but the color is the same as the shape color.
I think it's best to use the ShapeManager values to set for the comments in figure (i.e. controlling the appearance of the labels from figure, not from iviewer and refresh)

### DIFF
--- a/src/js/views/roi_loader_view.js
+++ b/src/js/views/roi_loader_view.js
@@ -93,8 +93,9 @@ var RoiLoaderView = Backbone.View.extend({
                         y: y,
                         textAnchor: "middle",  // center ROI labels by default
                         fillOpacity: 0,
+                        fillColor: "#ffffff",
                         strokeColor: shapeJson.strokeColor
-                        // fontSize and fillColor will be populated by the ShapeManager
+                        // fontSize will be populated by the ShapeManager
                     }
                     shapeJson.textId = textShape.id;  // link text shape to parent shape
                     shapeList.push(textShape)


Tom-TBT:
I implemented these change to display the shape comments.
* if ROI has a single shape, comment display in the ROI
* if there's more than one shape, then display only on the shape like in the iviewer
* if the comment is too long, the text is replaced by an ellipsis
* the tooltip now also shows the shape comment (in full)  → should it have an upper limit, like 100 chars?
 
image

I also fixed the anchor to ROI labels to "middle". 
Other properties are taken from the toolbar (textColor, backgroundOpacity, ...)

will-moore:
Thanks @Tom-TBT - I realised when testing that only the Rectangle handles the `textId`. I pushed a commit that adds it to the other shapes: https://github.com/will-moore/omero-figure/commit/090156fd7d8853fa176ac5103592c05a4bd6348f if you could cherry-pick that?

I also noticed that the colour of the new label is determined by the current status of the colour chooser, rather than the colour of the shape you're adding. This is a bit unexpected - while adding shapes, I clicked on one to select it (which also changed the colour selection widget) and then subsequent shapes got a different Text colour than before.
So I think that new labels should start off with the same colour as their parent shape, and transparent background (fill).

PR #657, opened 2025-12-18
Avoid division by zero and don't save x/y NaN in grid layout

will-moore:
Fix issue reported at https://github.com/ome/omero-figure/pull/651#discussion_r2631608389

To test:
 - Try various random / invalid / overlapping positions of panels and try to align to grid (e.g. see comment above)
 - The results may not be predictable (which is OK) but at least we avoid dividing by `0` (no Infinity and infinite loops - app becomes unresponsive!) and we also MUST avoid saving `NaN` to `x` or `y`.

cc @Rdornier


Rdornier:
Looks good to me, thanks @will-moore

PR #682, opened 2026-04-16
Enable pre-release rc workflow with bumpversion and publish to pypi

will-moore:
This PR adds pre-release support to bumpversion workflow to prevent the accidental publishing of rc versions as full releases.

Issue: Running:

```
bumpversion --new-version 8.0.0rc1 release
```
actually set the `__version__ = "8.0.0"` because "bumpversion config in .bumpversion.cfg:1 only models dev/prod, not rc. So your current automated bump flow does not support release candidates cleanly" (copilot).

See the resulting commit at https://github.com/ome/omero-figure/commit/3e44ff98c4f581f706313f45aa420e2231e06b92

Also, the `publish_pypi.yml` workflow didn't check that the gh tag and the `__version__` were the same.

All the changes in this PR are from github copilot. 

See `README` for details of how to do full release, specifying version (no change) and rc release.

To test:
 - checkout this branch
 - Run `bumpversion release` and check that the `__version__` in the commit
 - Then delete your local tag and reset to previous commit etc (don't push!)

When I try this, I see e.g. utils.py

```
-__version__ = "8.0.1.dev0"
+__version__ = "8.0.1.rc0"
```
And the Tag created is `tag: v8.0.1.rc0`.

<details><summary>Full github copilot discussion</summary>

Me: "I want to be able to create a pre-release tag that will publish a pre-release version to pypi. When I created a tag 8.0.0rc1 this published a full release 8.0.0 to pypi which is NOT what I want. Do I need to change the github workflow here, or use a different pre-release tag naming?"

Optimizing tool selection...

I’ll inspect the publish workflow and the package versioning logic to see why a tag like 8.0.0rc1 ended up as 8.0.0 on PyPI, then I’ll tell you whether the fix belongs in tag naming, workflow logic, or both.

I found the workflow always publishes on any tag; next I’m checking where the package version is sourced, since PyPI uses package metadata version (not tag text) unless you wire it in.

I’ve confirmed the key point: PyPI publication uses the package’s declared version from source, not the Git tag string. I’ll quickly check the release/version tooling config to confirm the intended version bump flow before giving a concrete recommendation.

I’ve isolated the cause and the safest fix path: your workflow publishes any tag, but the actual published version comes from source code version fields, so I’ll give you an exact release rule set plus an optional workflow guard to prevent this happening again.Use both: correct pre-release version in source and a small workflow guard. Tag naming alone is not enough.

What happened:

[...]

PR #684, opened 2026-04-20
Fix PKG_VERSION fix in publish_pypi.yml

will-moore:
This aims to fix the error at https://github.com/ome/omero-figure/actions/runs/24660519372/job/72105014376

Running locally, this works:

```
$ PKG_VERSION=$(python -c "import omero_figure.utils as u; print(u.__version__)")
$ echo $PKG_VERSION 
8.0.1.rc0
```

### DIFF
--- a/.github/workflows/publish_pypi.yml
+++ b/.github/workflows/publish_pypi.yml
@@ -11,36 +11,6 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: '3.10'
-      - name: Validate tag matches package version
-        if: startsWith(github.ref, 'refs/tags')
-        run: |
-          TAG="${GITHUB_REF#refs/tags/}"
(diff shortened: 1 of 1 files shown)


will-moore:
cc @jburel - seems I trusted the github copilot code here a bit too much.
I think the commit above should fix the syntax error. But I'm also thinking that maybe we don't need that whole build step to compare the version number in `figure.utils.py` with the github tag? If we stick to the correct `bumpversion` commands that we've tested then we should be good, so I think we could remove this whole `- name: Validate tag matches package version` step that was added in #682 ?

jburel:
Apologies I missed that check. 
We do not have that anywhere. We could certainly improve it in a follow up strategy.
Please remove

will-moore:
Done, removed that step from publish_pypi.yml

PR #692, opened 2026-06-16
colorbar lut background fix

will-moore:
Fixes the display of LUT for colorbar when deployed from OMERO.web.

To test:

 - Add a colorbar and set channel to LUT when deployed from OMERO
 - colorbar background should be rendered correctly as LUT


pwalczysko (approved):
wfm on merge-ci, thanks

Issue #699, opened 2026-08-11
Figure export to PDF errors on zoom level check


govekk:
https://github.com/ome/omero-figure/blob/ad6d35755a0406e60cf7c349dcb3a4a03f255dd5/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py#L2079-L2081

causes error:

```
Traceback (most recent call last):
  File "/home/omero-server/omero/tmp/omero_omero-server/1315/process89yj3uer.dir/./script", line 3248, in <module>
    run_script()
  File "/home/omero-server/omero/tmp/omero_omero-server/1315/process89yj3uer.dir/./script", line 3234, in run_script
    file_annotation = export_figure(conn, script_params)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/omero-server/omero/tmp/omero_omero-server/1315/process89yj3uer.dir/./script", line 3186, in export_figure
    return fig_export.build_figure()
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/omero-server/omero/tmp/omero_omero-server/1315/process89yj3uer.dir/./script", line 1219, in build_figure
    self.add_panels_to_page(panels_json, image_ids, page)
  File "/home/omero-server/omero/tmp/omero_omero-server/1315/process89yj3uer.dir/./script", line 2515, in add_panels_to_page
    image, pil_img = self.draw_panel(panel, page, i)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/omero-server/omero/tmp/omero_omero-server/1315/process89yj3uer.dir/./script", line 2336, in draw_panel
    pil_img = self.get_panel_image(image, panel, orig_name)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/omero-server/omero/tmp/omero_omero-server/1315/process89yj3uer.dir/./script", line 2240, in get_panel_image
    pil_img = self.get_panel_big_image(image, panel)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/omero-server/omero/tmp/omero_omero-server/1315/process89yj3uer.dir/./script", line 2176, in get_panel_big_image
    pil_img = self.render_big_image_region(image, panel, z, t,
[...]

PR #700, opened 2026-08-13
Fix parentheses in zoom level for loop for export script

### LINKED ISSUE #604: Export issue when the lowest resolution is greater than the max plane size
The issue can be simulated with a synthetic image with the following dimensions

```bash
$ cat test.fake.ini 
sizeX=100000
sizeY=100000
resolutions=5
```

After importing this image into OMERO, creating a figure and trying to export it, the script fails with the following error:

```
  File "./script", line 2562, in <module>
    run_script()
  File "./script", line 2548, in run_script
  File "./script", line 2504, in export_figure
  File "./script", line 980, in build_figure
    self.add_panels_to_page(panels_json, image_ids, page)
  File "./script", line 1988, in add_panels_to_page
    image, pil_img = self.draw_panel(panel, page, i)
  File "./script", line 1809, in draw_panel
    pil_img = self.get_panel_image(image, panel, orig_name)
  File "./script", line 1707, in get_panel_image
    pil_img = self.get_panel_big_image(image, panel)
  File "./script", line 1643, in get_panel_big_image
    pil_img = self.render_big_image_region(image, panel, z, t,
  File "./script", line 1563, in render_big_image_region
    scale, level = self.get_zoom_level_scale(image, region, max_width)
  File "./script", line 1548, in get_zoom_level_scale
    zm_levels[zm] * width * zm_levels[zm] * height > max_plane):
KeyError: 5
```
The logic in
https://github.com/ome/omero-figure/blob/1dfb1389e449594defce3d4ac9f72df3fefb6e89/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py#L1700-L1704
assumes that a resolution lower than the maximum plane size will be found which is incorrect. Minimally this should be updated  to handle the case where the lowest resolution is greated than 3kx3k in a default configuration.

This raises the question of the how the export script and more generally the application should behave in these scenarios. One approach would be to use tile render calls rather than plane calls but this might be a larger effort. Another option would be to block render calls with an informative error message when the lowest resolution is larger than the maximum plane size  similarly to what was done in https://github.com/ome/omero-figure/pull/560 for maximum projection.

will-moore:
Fixes #604 and #699.

The applies the fix identified in both of those issues.

To test, create a big image with limited resolutions, e.g.

`pyramid&sizeX=20000&sizeY=20000&resolutions=3.fake`

Import this image into OMERO and add to a Figure.
See https://merge-ci.openmicroscopy.org/web/webclient/?show=image-2215 
Try to export the Figure. Without this PR, you will see the`KeyError` reported in issues above. 

With this PR, the Figure should export as normal

### DIFF
--- a/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
+++ b/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
@@ -2165,8 +2165,8 @@ class FigureExport(object):
         zm = 0
         max_plane = max_sizes[0] * max_sizes[1]
         while (zm < max_level and
-               zm_levels[zm] * width > max_width or
-               zm_levels[zm] * width * zm_levels[zm] * height > max_plane):
+               (zm_levels[zm] * width > max_width or
+                zm_levels[zm] * width * zm_levels[zm] * height > max_plane)):
             zm = zm + 1
 
         level = max_level - zm


pwalczysko (approved):
wfm on the merge-ci (with this PR).
without this PR, I get

```
  File "/home/omero-x/omero/tmp/omero_omero-x/5127/processf7m3o34t.dir/./script", line 3248, in <module>
    run_script()
  File "/home/omero-x/omero/tmp/omero_omero-x/5127/processf7m3o34t.dir/./script", line 3234, in run_script
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/omero-x/omero/tmp/omero_omero-x/5127/processf7m3o34t.dir/./script", line 3186, in export_figure
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/omero-x/omero/tmp/omero_omero-x/5127/processf7m3o34t.dir/./script", line 1219, in build_figure
    self.add_panels_to_page(panels_json, image_ids, page)
  File "/home/omero-x/omero/tmp/omero_omero-x/5127/processf7m3o34t.dir/./script", line 2515, in add_panels_to_page
    image, pil_img = self.draw_panel(panel, page, i)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/omero-x/omero/tmp/omero_omero-x/5127/processf7m3o34t.dir/./script", line 2336, in draw_panel
    pil_img = self.get_panel_image(image, panel, orig_name)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/omero-x/omero/tmp/omero_omero-x/5127/processf7m3o34t.dir/./script", line 2240, in get_panel_image
    pil_img = self.get_panel_big_image(image, panel)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/omero-x/omero/tmp/omero_omero-x/5127/processf7m3o34t.dir/./script", line 2176, in get_panel_big_image
    pil_img = self.render_big_image_region(image, panel, z, t,
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/omero-x/omero/tmp/omero_omero-x/5127/processf7m3o34t.dir/./script", line 2096, in render_big_image_region
    scale, level = self.get_zoom_level_scale(image, region, max_width)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/omero-x/omero/tmp/omero_omero-x/5127/processf7m3o34t.dir/./script", line 2081, in get_zoom_level_scale
    zm_levels[zm] * width * zm_levels[zm] * height > max_plane):
    ~~~~~~~~~^^^^
KeyError: 3
```

sbesson:
While this PR addresses an incorrect conditional statement and prevents the export script to error with a cryptic `KeyError`, have the questions raised in the last paragraph of https://github.com/ome/omero-figure/issues/604#issue-2720063031 been considered during implementation and/or the review?

The figure export script is implemented as a series of rendering requests to the server. As this fix specifically images where the lowest resolution is exceeds the maximum plane width/height, what mechanisms are put into place to ensure the dimensions of the requested tiles still meet reasonable requirements?

will-moore:
@sbesson Since a couple of users have reported that the logic fix in this PR allows them to export figures that previously failed with `KeyError`, the priority was to get this simple fix into the next release.

The issue of dealing with "smallest pyramid resolution is too big" is much harder and is not even limited to OMERO.figure.
There is no defined "maximum plane width/height". We could use https://omero.readthedocs.io/en/stable/sysadmins/config.html#omero-client-download-as-max-size as the threshold, or possibly a bit smaller.

We would also need to duplicate the logic in the omero-web code ([render_scaled_region](https://github.com/ome/omero-figure/blob/81f95f4400b3cda77eae8fa9ea94071a1253a876/omero_figure/views.py#L315)) as well as in the Figure_To_Pdf.py script. There is a good argument for putting that logic into omero-web or even omero-py.

So, this is all doable but just hasn't been a priority yet.

will-moore:
The plan is to release the ome-zarr functionality (probably in `v8.0.2`) ASAP and then focus on splitting the JS app into a new `ome/figure` repo.

sbesson:
> The issue of dealing with "smallest pyramid resolution is too big" is much harder and is not even limited to OMERO.figure.

Absolutely, addressing the problem of insufficient resolution levels fully will require to reconsider the entire pipeline starting from data import.

Independently of this work, clients & applications have some responsibility when making requests in a multi-user scenario. This is probably even more critical in OMERO.figure given its prominence in the OMERO ecosystem. https://github.com/ome/omero-figure/pull/560 was a good recent example of implementing similar guardrails in a different context.

> There is no defined "maximum plane width/height".

I disagree here. `omero.pixeldata.max_plane_height` and `omero.pixeldata.max_plane_width` are two long defined concepts which specify maximum plane dimensions above which data should no longer treated and accessed as a single-resolution whole planes.

Although there is no enforcement at the server level,  these values are used in a number of place as a limit for whole plane access - see  https://github.com/ome/omero-server/pull/162 for a recent example.

jburel:
Catching up with this PR, I completely missed it. 

I agree that this should be done carefully. We have seen similar problems for the projection. It could kill the server without clear assessment of the impact.

It is dangerous to squeeze this PR in before the split. The change is unrelated to the split. 
Proposal:
* Revert the commit before release.

PR #704, opened 2026-08-21
Revert commit 49ae9c9

jburel:
This PR reverts a commit introduced in #700.
The change required an in-depth review since it could have significant impact on the server.
This change is not needed before the split.

### DIFF
--- a/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
+++ b/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
@@ -2165,8 +2165,8 @@ class FigureExport(object):
         zm = 0
         max_plane = max_sizes[0] * max_sizes[1]
         while (zm < max_level and
-               (zm_levels[zm] * width > max_width or
-                zm_levels[zm] * width * zm_levels[zm] * height > max_plane)):
+               zm_levels[zm] * width > max_width or
+                zm_levels[zm] * width * zm_levels[zm] * height > max_plane):
             zm = zm + 1
 
         level = max_level - zm


will-moore:
Let's not "fix" this issue by adding back a bug.
That bug gives a non-informative `KeyError` and the whole export script fails.

Instead, to ensure that we don't attempt to render images larger than `omero.pixeldata.max_plane_height` and `omero.pixeldata.max_plane_width`, let's check the dimensions of the chosen resolution and if the limit is exceeded then the export script can simply skip that particular panel.

jburel:
The comment in this original PR says: ```With this PR, the Figure should export as normal.``` unless I am missing something, it does not fail the export.
Happy if you take that over, as it stands we can't release without a clear assessment of  the impact if the original commit is included.

govekk:
Popping in as an OMERO admin who reported the bug: I agree with Will that the solution can't be to add back incorrect logic with a non-informative error. If you wanted exactly the same behavior, you could check if the requested rendering size is lower than the lowest pyramid available and error out with an informative error. Though I prefer Will's option of checking against a config setting, so that admins can decide to allow rendering down of some large images - that's a common need with the WSI on our server.

jburel:
I don't disagree that this is not the best way (I do not have the time to look at it).  It is just that the commit "fixing" the bug could lead to major issue server side. This needs to be fixed before we release.

will-moore:
Closing in favour of #705.

will-moore:
Does this need testing, or are we OK since we're simply reverting back to previous code?

jburel:
testing using the large fake file
```
  File "/home/omero/omero/tmp/omero_omero/152427/process2khiyqq2.dir/./script", line 3562, in <module>
    handle_main()
  File "/home/omero/omero/tmp/omero_omero/152427/process2khiyqq2.dir/./script", line 3506, in handle_main
    run_script()
  File "/home/omero/omero/tmp/omero_omero/152427/process2khiyqq2.dir/./script", line 3485, in run_script
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/omero/omero/tmp/omero_omero/152427/process2khiyqq2.dir/./script", line 3437, in export_figure
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/omero/omero/tmp/omero_omero/152427/process2khiyqq2.dir/./script", line 1257, in build_figure
    self.add_panels_to_page(panels_json, page)
  File "/home/omero/omero/tmp/omero_omero/152427/process2khiyqq2.dir/./script", line 2761, in add_panels_to_page
    pil_img = self.draw_panel(panel, page, i)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/omero/omero/tmp/omero_omero/152427/process2khiyqq2.dir/./script", line 2580, in draw_panel
    pil_img = self.get_panel_image(panel, orig_name)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/omero/omero/tmp/omero_omero/152427/process2khiyqq2.dir/./script", line 2471, in get_panel_image
    pil_img = self.get_panel_big_image(panel)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/omero/omero/tmp/omero_omero/152427/process2khiyqq2.dir/./script", line 2297, in get_panel_big_image
    pil_img = self.render_big_image_region(panel, z, t,
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/omero/omero/tmp/omero_omero/152427/process2khiyqq2.dir/./script", line 2226, in render_big_image_region
    scale, level = self.get_zoom_level_scale(image, region, max_width)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/omero/omero/tmp/omero_omero/152427/process2khiyqq2.dir/./script", line 2169, in get_zoom_level_scale
    zm_levels[zm] * width * zm_levels[zm] * height > max_plane):
    ~~~~~~~~~^^^^
KeyError: 5
```
As expected

PR #705, opened 2026-08-25
If big image region is too big, don't try to render

will-moore:
Follow-up of #700.

With the fix of the logic for picking zoom levels in #700, it is possible that the export script attempts to render a region larger than `max plane size` (`omero.pixeldata.max_plane_width * omero.pixeldata.max_plane_height`) if the image's smallest pyramid resolution is still quite large.

This PR adds an extra check to make sure the chosen resolution doesn't exceed that size limit before we try to render.

If it does, we simply skip the addition of that panel to the figure and the rest of the figure should be exported as normal.

To test:

 - create / find big image with few resolutions (as described in #700)
 - Add image to figure (along with other images) and export the figure
 - The offending image should be skipped and the rest of the figure exported as normal

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -70,6 +70,13 @@ Optional: To change the maximum active channel count from the default of 10:
 
     $ omero config set omero.figure.max_active_channels 15  
 
+Optional: Big images, where the panel viewport at lowest pyramid resolution exceeds the "big image" threshold
+(approx 3000 x 3000 pixels) are excluded from the exported figures.
+To set a different threshold, e.g. 5000 x 5000 pixels, you can do:
+
+::
+
+    $ omero config set omero.figure.max_panel_pixels 25000000
 
 Now restart OMERO.web as normal.
 
--- a/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
+++ b/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
@@ -1059,6 +1059,11 @@ class FigureExport(object):
 
         self.conn = conn
         self.script_params = script_params
+        self.max_panel_pixels = script_params.get("Max_Panel_Pixels")
+        if self.max_panel_pixels is None:
+            if conn is not None:
+                max_plane_size = self.conn.getMaxPlaneSize()
+                self.max_panel_pixels = max_plane_size[0] * max_plane_size[1]
         self.export_images = export_images
         self.errors = []
         # For standalone script, we may have relative or absolute output path
@@ -2237,21 +2242,19 @@ class FigureExport(object):
             size_y = int(size_y * scale)
 
             # check if region is larger than max plane size...
-            max_sizes = self.conn.getMaxPlaneSize()
-            if width * height > max_sizes[0] * max_sizes[1]:
+            if width * height > self.max_panel_pixels:
                 self.errors.append(
                     f"Failed to render image {image.name} (ID: {image.id}):" +
                     " Smallest pyramid resolution is too large to render:" +
-                    f" {width} x {height} > {max_sizes[0]} x {max_sizes[1]}")
+                    f" {width} x {height} > {self.max_panel_pixels}")
                 # return a placeholder PIL Image with "failed to render" text
                 ph_width = 400
                 ph_height = int(ph_width * (height / width))
-                font = get_font(30)
                 pil_img = Image.new("RGBA", (ph_width, ph_height),
                                     (221, 221, 221, 255))
                 draw = ImageDraw.Draw(pil_img)
                 draw.text((10, 10), "Failed to render",
-                          fill=(255, 0, 0, 255), font=font)
+                          fill=(255, 0, 0, 255), font=get_font(30))
                 return pil_img
 
             try:
@@ -3469,6 +3472,10 @@ def run_script():
 
         scripts.String("Figure_URI",
                        description="URL to the Figure"),
+
+        scripts.Long("Max_Panel_Pixels",
+                     # default limit is conn.getMaxPlaneSize()
+                     description="Don't render panels with image-regions exceeding this limit"),
 
         # This allows clients to query the script version
         # by listing script params and getting default value
--- a/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
+++ b/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
@@ -1059,6 +1059,7 @@ class FigureExport(object):
         self.conn = conn
         self.script_params = script_params
         self.export_images = export_images
+        self.errors = []
         # For standalone script, we may have relative or absolute output path
         self.output_path_name = script_params.get("outputPathName")
 
@@ -2237,9 +2238,15 @@ class FigureExport(object):
             # check if region is larger than max plane size...
             max_sizes = self.conn.getMaxPlaneSize()
             if width * height > max_sizes[0] * max_sizes[1]:
-                print("Region is too large to render: %d x %d > %d x %d" %
-                      (width, height, max_sizes[0], max_sizes[1]))
-                return None
+                self.errors.append(
+                    f"Failed to render image {image.name} (ID: {image.id}):" +
+                    f" Smallest pyramid resolution is too large to render:" +
+                    f" {width} x {height} > {max_sizes[0]} x {max_sizes[1]}")
+                # return a placeholder PIL Image with "failed to render" message 
+                pil_img = Image.new("RGBA", (width//10, height//10), (221, 221, 221, 255))
+                draw = ImageDraw.Draw(pil_img)
+                draw.text((10, 10), "Failed to render", fill=(255, 0, 0, 255))
+                return pil_img
 
             try:
                 self.apply_rdefs(image, panel)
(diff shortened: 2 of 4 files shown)


will-moore:
One concern I have with the limit added in this PR is that it is quite conservative - e.g. typically `3k * 3k` when we know that the server will easily be able to render e.g. 5k * 5k or more without issues.

But because we use the `omero.pixeldata.max_plane_width * omero.pixeldata.max_plane_height` settings here, it's not easy for a user to configure a slightly higher limit, since these values are used to determine the "Big Image" threshold for the server (am I correct about that)?

A more suitable threshold that is designed specifically for the rendering of larger images for export is the [omero.client.download_as.max_size](https://omero.readthedocs.io/en/stable/sysadmins/config.html#omero-client-download-as-max-size) setting. The default is `144000000` (12k * 12k) and it's used to determine whether you `canExportAsJpg()` for an image in the download menu of the webclient.

will-moore:
@govekk - for your images where you've seen this issue in omero-figure, what size is the lowest resolution? Would the 3k x 3k limit exclude them from being rendered in the PDF? If they are less than 12k x 12k, do you see any performance / memory issues when exporting figures with these images?

will-moore:
An alternative option is that we create a *new* omero-figure setting for the max rendering size threshold. The default value for this could be fairly conservative. e.g. 4k * 4k, but this would allow users to configure this in case they need to?

jburel:
I tested the PDF generation by creating an image of the size described in #604.
There is no crash so the user has the impression that everything went smoothly but there is actually no image in the PDF see pdf attached.
[Figure_2026-9-1_13-55-38.pdf](https://github.com/user-attachments/files/31695293/Figure_2026-9-1_13-55-38.pdf)

will-moore:
@jburel That is working exactly as described above.
I see this as a very edge case handling here - we are simply avoiding the rare case of trying to render an image that has it's lowest resolution too big.
You didn't say how you'd want the behaviour to change?

jburel:
I understand the description of the PR, I should have indicated that I put myself as "user".
Selecting one big image and skipping it without any notification is not what I will expect. 
The impression is that "everything works" but in reality nothing happened. Having the skipped image in the list of  images "composing" the figure adds even more confusion. 

The user will expect that the strategy used to render the image in the "web" figure is also used to generate the PDF content

* A report should be returned to the user so they are aware that an image is not included.
* The skipped image should not be included in the list of images composing the figure

govekk:
> @govekk - for your images where you've seen this issue in omero-figure, what size is the lowest resolution? Would the 3k x 3k limit exclude them from being rendered in the PDF? If they are less than 12k x 12k, do you see any performance / memory issues when exporting figures with these images?

Apologies for the wait: the image that triggered this issue for us has a lowest res of 12k x 5k (I think due to some issue with the files actually, that's a separate thing I'm tracking down). So it wouldn't fit under the 3k x 3k limit but I think it's reasonable to enforce such a limit, though it would be great as an admin to have control over that based on the resources you can give your server.

When exporting the 12k x 5k res image with the fix I suggested in #699, the script reaches about 3Gb mem usage max. It seems to be single-threaded so it's not really maxing out cpu or mem, but I could see that being some pressure on a smaller server.

will-moore:
Comparing parameters: add print statements in views.py and export script to compare values passed to `renderJpegRegion()`.

```
+++ b/omero_figure/views.py
@@ -372,6 +372,7 @@ def render_scaled_region(request, iid, z, t, conn=None, **kwargs):
             y = 0
 
     # Render the region...
+    print("renderJpegRegion: x=%s, y=%s, width=%s, height=%s, level=%s, compress_quality=%s" % (x, y, width, height, level, compress_quality))
     jpeg_data = image.renderJpegRegion(z, t, x, y, width, height, level=level,
                                        compression=compress_quality)
```
and
```
+++ b/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
@@ -2250,6 +2250,7 @@ class FigureExport(object):
 
             try:
                 self.apply_rdefs(image, panel)
+                print("renderJpegRegion: x=%s, y=%s, width=%s, height=%s, level=%s" % (x, y, width, height, level))
                 jpeg_data = image.renderJpegRegion(z, t, x, y, width, height,
                                                    level=level)
                 if jpeg_data is None:
```

Testing with a fake image named: `pyramid1&sizeX=50000&sizeY=50000&resolutions=3.fake` which has following resolutions:

```
"resolutions": {
        "0": {
        "sizeX": 50000,
        "sizeY": 50000
    },
        "1": {
        "sizeX": 25000,
        "sizeY": 25000
    },
        "2": {
        "sizeX": 12500,
        "sizeY": 12500
    }
},
```

NB: when rendering Full Size...

Web:
```
/Users/wmoore/opt/anaconda3/envs/omeroweb_zarrv3/lib/python3.12/site-packages/PIL/Image.py:3442: DecompressionBombWarning: Image size (156250000 pixels) exceeds limit of 89478485 pixels, could be decompression bomb DOS attack
```
script:
```
/opt/omero/server/venv3/lib/python3.10/site-packages/PIL/Image.py:3167: DecompressionBombWarning: Image size (156250000 pixels) exceeds limit of 89478485 pixels, could be decompression bomb DOS attack.
```

Fully zoomed out, web then script:
```
renderJpegRegion: x=0, y=0, width=15625, height=15625, level=0, compress_quality=None
renderJpegRegion: x=0, y=0, width=12500, height=12500, level=0
```

Zoom to 1000%, web then script:
```
renderJpegRegion: x=5312, y=5312, width=1875, height=1875, level=0, compress_quality=None
renderJpegRegion: x=11250, y=11250, width=2500, height=2500, level=1
```

Zoom to 2000%, web then script:
```
renderJpegRegion: x=11562, y=11562, width=1875, height=1875, level=1, compress_quality=None
enderJpegRegion: x=23750, y=23750, width=2500, height=2500, level=2
```

Zoom to 5000%,
```
renderJpegRegion: x=24250, y=24250, width=1500, height=1500, level=2, compress_quality=None
renderJpegRegion: x=24500, y=24500, width=1000, height=1000, level=2
```

will-moore:
@jburel So, we get slightly different thresholds for picking resolutions for web & script when we partially zoom in to the image, but when fully zoomed out or zoomed in a long way then we get the same resolution level chosen for both.

The web rendering picks a larger area (1.5 times the viewport region) to give extra for panning the image in the app.

will-moore:
The PDF export now shows a placeholder image with "Failed to render" and the error text is shown in the info page:

Screenshot 2026-09-02 at 18 11 47

jburel:
Thanks for in the investigation. The "level" does not match in some cases, so users don't add to the pdf when they see in the web. @sbesson @knabar do you know if that is the case where the OOM exception was noticed.

sbesson:
I agree the investigation of the rendering parameters for different zooming levels is useful. I don't think we have comparable measurements. If these parameters are useful to review some of these decisions, would it make sense to log them as DEBUG/INFO statements?

Many thanks to @govekk for collecting and reporting metrics in https://github.com/ome/omero-figure/pull/705#issuecomment-5512077332. They are fully relevant and support the concerns raised in https://github.com/ome/omero-figure/pull/700. Having a single figure export operation requiring multiple GBs of memory might be acceptable in some situations like a single-user environment but can have real impact in a heavily multi-user context especially alongside other memory intensive operations like imports. Assuming these figures were generated from RGB WSI data, I expect the required memory can increase by several factors when dealing with multiplexed  images or data types like floating-point.

These are the reasons why we proposed to reduce the maximum requested plane size from 12kx12k to the standard 3kx3k in https://github.com/ome/omero-figure/pull/553. I don't think we want to change this default without additional data points although we might want to make it configurable as discussed above so that an OMERO administrator might elect to assign more resources to the OMERO.figure export workflow

will-moore:
@sbesson One question that @jburel and I were discussing yesterday is why we haven't seen memory issues when rendering big images in the figure app? E.g. in the example above, if your lowest pyramid resolution is 12500 x 12500 and you're able to view that in the web app, that whole plane is being rendered by `renderJpegRegion()` in views.py, just the same as in the export script.
Should we allow larger planes to be rendered in the web app views.py than we allow in the export script?
This is the current behaviour. It allows you to construct your figure and view it in the web (which is nice), but with this PR you won't be able to render some panels in the export script.

will-moore:
OK, so those last commits add the option to do this:  (added to README)

```
$ omero config set omero.figure.max_panel_pixels 25000000
```

Without that, the limit stays at the "big image" 3k x 3k threshold. Images bigger than that will be skipped with error as in screenshot above.

will-moore:
Current status:

 - In the zoom levels logging above, you can see that the area rendered (width x height passed to renderJpegRegion()) is very much larger when fully zoomed out: e.g. `12500 x 12500` at level 0 compared with `2500 x 2500` at level 1 when zoomed in 10x. So, the rendering load (and risk of OOM) is much higher when fully zoomed out.
 - Also, when fully zoomed out, the lowest resolution level is correctly chosen in both the web and python script usage.
 - We haven't been able to reproduce an OOM in our testing so we can't test directly...
 - ...but, to be safe, this PR introduces a limit of 3k x 3k. Above this limit we show a placeholder image and display more complete error message in the PDF info page
 - That limit is configurable: `$ omero config set omero.figure.max_panel_pixels 25000000`

Next steps:

 - Test image [pyramid1&sizeX=100000&sizeY=100000&resolutions=5.fake](https://merge-ci.openmicroscopy.org/web/webclient/?show=image-11465) ID: 11465 on merge-ci had dimensions and resolutions as Seb tested with earlier. https://merge-ci.openmicroscopy.org/web/webclient/imgData/11465/ shows that lowest resolution is `6250 x 6250`.
 - Testing of this PR by choosing a big image with limited resolutions (lowest resolution over 3k x 3k), adding it to a figure and checking that it is excluded from exported PDF/TIFF (as in screenshots above).
 - Test config of a higher limit, e.g. 10k x 10k. (without any limit - before this PR - I have failed to get an OOM with multiple 12k x 12k images).

jburel:
Thanks for the report. Commit https://github.com/ome/omero-figure/commit/49ae9c9aa9514f87fbb0a2f5e03d5408e240036d introduced a "quick" fix that was not planned and not fully tested. It also highlights a larger problem that required careful consideration, discussion and work (also some commits in this PR mean than this is no longer a patch release).

This PR is currently preventing the split (see https://github.com/ome/omero-figure/pull/706), which was the original plan. 
Post split, rendering and export will need to happen in the browser and not the server. So major rework is required regardless.

I understand the frustration this might cause but, in order to move forward with the plan, we have to:
* Revert the commit
* Close this PR
* Tag
* Start the split

